PR #50, opened 2014-04-16
add optional bilingual setup

maxheld83:
Here's my hopefully not too crufty bilingual setup for minimal mistakes.

It leaves everything untouched as far as possible, and people can just delete the multilingual post and pages, and everything should look normal.

This setup allows:
- translated posts and pages as well as those that are available only in one language
- navbar, bio, footer etc. adapt to the chosen language
- post indices will include only the posts in the language of the `post.md`
- all necessary metadata (`lang: de`) is simply declared in the frontmatter.
- links (`trans: ...`) to translations are "hard-coded".
- some very few site-wide fields needed to be adapted, including `bio`and `links`; these now feature sub items according to JSON. The defaults are preserved as `en`
- the language selector shows up only if there is something to be translated. On non-translated posts or pages, the site appears to be monolingual. 

To dos:
- [ ] the language selector looks pretty bad. I just somehow wedged it into the navbar. Any better design ideas, @mmistakes? Maybe the top right corner?

I understand that it's important to keep minimal-mistakes nice and, well, _minimal_ and maybe this setup introduces too much complexity.

Still, I think a lot of people need a bilingual setup, and might use this as a starting point and/or make minimal-mistakes stand out as one of the few themes with (sorta) out-of-the-box bilingual support.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -8,7 +8,9 @@ disqus_shortname:
 owner:
   name:           Your Name
   avatar:         bio-photo.jpg
-  bio:            "Describe your self."
+  bio:            
+    en: "Describe your self."
+    de: "Beschreiben Sie sich."
   email:          your@email.com
   # Social networking links are used in author-bio sidebar. Update and remove as you like.
   twitter:        


rriemann:
another solution for mullti-lingual content: https://github.com/untra/polyglot

Issue #113, opened 2014-12-22
Setting url in _config.yml yields invalid resource urls over localhost


ottumm:
If I set `url` in `_config.yml` to my site's public domain, then css, js, and image files fail to load when running over localhost (via `jekyll serve`). However, if I don't set `url` then the home navigation header (top left) doesn't work.

I may be doing something wrong in my dev environment. In any case, I appreciate the help!

PR #114, opened 2014-12-22
Use relative urls for most internal site links

ottumm:
This allows both development and production instances to work in the same way. Links within the site begin with '/' rather than beginning with {{ site.url }}.

Fixes issue #113.

(I understand if this pull request goes against the design philosophy of the project. If that's the case, please feel free to close the request or disregard it.)


RamblingCookieMonster:
Just went through this exercise as well, folks with HTTPS everywhere were complaining of a 90's style webpage : )  Would definitely recommend moving to relative paths where it makes sense.

Cheers!

mmistakes:
@RamblingCookieMonster If you remove the prototcol from `{{ site.url }}` it'll work for both `http` and `https`

```
url: //yourdomain.com
```

The only other issue that may come up if you care about SEO is duplicate content in the eyes of Google. I don't think there is a way to force redirect http traffic to https on sites hosted on GitHub Pages which is usually the preferred way around this. But I guess that's up to user and out of scope of the theme.

PR #143, opened 2015-04-30
add reading time on post and page

mastito03:


prakash014:
:+1: Great feature @mastito03 ! 
I used it in my blog and loved it.

**Suggestion:** I don't think this feature is needed in `page`, hope @mmistakes merges this feature soon.

PR #146, opened 2015-05-22
add RSS icon

srijanshetty:
I've added an RSS icon in the author bio, found it kinda handy

### DIFF
--- a/_includes/_author-bio.html
+++ b/_includes/_author-bio.html
@@ -32,3 +32,4 @@
   {% if author.weibo %}<a href="http://www.weibo.com/{{ author.weibo }}" class="author-social" target="_blank"><i class="fa fa-fw fa-weibo"></i> Weibo</a>{% endif %}
   {% if author.flickr %}<a href="http://www.flickr.com/{{ author.flickr }}" class="author-social" target="_blank"><i class="fa fa-fw fa-flickr"></i> Flickr</a>{% endif %}
 </div>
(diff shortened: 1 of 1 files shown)


prakash014:
@srijanshetty i dont think your code works!

It should be: `{{ site.url }}/feed.xml` instead of `{{ site.url/feed.xml }}`.

_includes/_author-bio.html
```diff
@@ -32,3 +32,4 @@ <h3 itemprop="name">{{ author.name }}</h3>
   {% if author.weibo %}<a href="http://www.weibo.com/{{ author.weibo }}" class="author-social" target="_blank"><i class="fa fa-fw fa-weibo"></i> Weibo</a>{% endif %}
   {% if author.flickr %}<a href="http://www.flickr.com/{{ author.flickr }}" class="author-social" target="_blank"><i class="fa fa-fw fa-flickr"></i> Flickr</a>{% endif %}
 </div>
+<a href="{{ site.url/feed.xml }}" class="author-social" target="_blank"><i class="fa fa-fw fa-rss"></i> RSS </a>
```
prakash014: @srijanshetty i dont think your code works!

It should be: {{ site.url }}/feed.xml instead of {{ site.url/feed.xml }}.

mmistakes:
As noted by @prakash014 to link to the site's feed it should be `{{ site.url }}/feed.xml`. That said not sure how I feel about merging this in. The sidebar is for **author** related links. By default a link in the footer or possibly the top menu seems like a better home for it.

It's also the kind of thing not everyone will want to feature in the sidebar, so it would need to be optional for me to consider it.

Issue #149, opened 2015-06-11
facebook share posts didn't pick the banner feature image.


kishore-mohan:
When we share the posts or any page into facebook. It's not sharing with banner image of the page, rather picking owner bio picture. In _open-graph.html for og-image meta tag content missing.

mmistakes:
Fixed in #150

PR #150, opened 2015-06-11
facebook sharing image meta tag added related #149 issue

kishore-mohan:
facebook meta tag image sharing added in the_open-graph.html realted to #149 issue. While sharing posts or page using facebook shar its not picking the rite image. Now its corrected adding ioen graph image tag added.


mmistakes:
Thanks for the PR.

Issue #153, opened 2015-06-19
Google+ Sharing button ignores OpenGraph info


ahrenstein:
Just launched my site with this theme last night. Everything is fantastic except that sharing with the Google+ button seems to ignore all of the OpenGraph info so the shared link reveals nothing other than the site name. Is there a good way to fix this?

Thanks

mmistakes:
Not entirely sure what's going on. As far as I can tell the theme is doing everything right as per Google.

> If the page contains Open Graph properties for the title, image, and description, they will be used for the +Snippet.

Which are included in the `<head>` of each page.

```
<meta property="og:title" content="..." />
<meta property="og:image" content="..." />
<meta property="og:description" content="..." />
```

Not sure why Google is ignoring it in the snippet. I tested sharing one of theme's demo pages on Google+ and it pulls in the feature image, but seems to use the author name as page title and doesn't add the description. No idea why it's doing that because the OG data seems accurate to me.

ahrenstein:
Hmmm so it's not just my site then. I'll keep looking into it as well.

mmistakes:
Possible solution would be to update the post/page layouts to [include schema.org microdata](https://developers.google.com/+/web/snippet/) to help identify what content is what. Back in the day I included a lot of these tags in the markup but took them out because it over complicated things and it didn't really fit all use cases.

Probably wouldn't be too hard for you to add it to your own fork if that's something you're comfortable doing.

ahrenstein:
Hmm I tried adding these tags to _includes/_open-graph.html but didn't see any changes

```
<meta itemscope itemtype="http://schema.org/Article" />
<meta itemprop="headline" content="{% if page.title %}{{ page.title }}{% else %}{{ site.title }}{% endif %}" />
<meta itemprop="description" content="{{ page.excerpt | strip_html }}" />
<meta itemprop="image" content="{% if page.image.thumb %}{{ site.url }}/images/{{ page.image.thumb }}{% else %}{{ site.url }}/images/default-thumb.png{% endif %}" />
```

ahrenstein:
@mmistakes could this issue be due to the aspect ratio of the image presented to Google+? https://developers.google.com/+/web/snippet/article-rendering

mmistakes:
Could be. There's nothing stopping you from using a taller image if you
wanted to test that out. You could also modify the theme to use a different
image for the OG one if the feature image is giving you problems. I
purposely kept things simple knowing most people won't even assign a
[...]

PR #154, opened 2015-06-19
Make social sharing a 600x600 pop-up

ahrenstein:
Changing social sharing buttons to provide a 600x600 pop-up instead hijacking the entire window will keep the visitor from leaving the site. Seems like better UX to me


ahrenstein:
Added another commit to fix the Schema.org and G+ issues mentioned in issue #153

PR #156, opened 2015-07-12
Add CodePen to owner social options

hbarudin:
I wanted to include CodePen on my site, and since there's already a [FontAwesome icon for it](http://fortawesome.github.io/Font-Awesome/icon/codepen/), seemed like an easy addition that I thought others could benefit from.


mmistakes:
Thanks!

PR #199, opened 2015-11-29
Fix home markup for link posts

kamilszymanski:


mmistakes:
Thanks for the PR @kamilszymanski

PR #229, opened 2016-02-17
Use uglifyjs npm script

timwis:
You can run this via `npm run build`. If you like it, you can remove the grunt-related dependencies from package.json. You can use the standalone jshint library if you want to keep linting as part of the project. You can do the same thing for image/svg compression.

You can create separate npm scripts in package.json and then run them with one command like this:
```json
{
  "lint": "jshint ./",
  "compress": "compressor ./img...",
  "uglify": "uglifyjs ...... -o ...",
  "build": "npm run lint && npm run compress & npm run uglify"
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
     "type": "git",
     "url": "git://github.com/mmistakes/minimal-mistakes.git"
   },
+  "scripts": {
+    "build": "uglifyjs assets/js/plugins/*.js assets/js/_*.js -o assets/js/scripts.min.js"
+  },
   "bugs": {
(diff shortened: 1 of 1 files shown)


mmistakes:
I could probably kill Grunt completely and just use [`npm scripts`](https://css-tricks.com/why-npm-scripts/) to do everything (watch, image compression, JS linting and uglifying, etc.). Something I'll have to think about.

timwis:
Indeed. Check out https://medium.com/@housecor/why-i-left-gulp-and-grunt-for-npm-scripts-3d6853dd22b8

PR #238, opened 2016-02-21
Feature: Move sample posts, pages, collections, and assets out of master

mmistakes:
- [ ] Remove sample posts, pages, collections, data files, and other assets from `master` branch to keep things nice and tidy.

Break out into their own repo?


mmistakes:
mm-layout-examples

PR #299, opened 2016-05-03
Add Speaker Deck author profile link

glennsarti:
This commit adds the Speaker Deck link to the author profile.  As there is no logo for Speaker Deck in the Font, a simple external link icon is used.

### DIFF
--- a/assets/_scss/_utilities.scss
+++ b/assets/_scss/_utilities.scss
@@ -258,6 +258,10 @@ body:hover .visually-hidden button {
     color: $stackoverflow-color;
   }
 
+  .fa-speaker-deck {
+    #color: $speakerdeck-color;
+  }
+
   .fa-tumblr,
   .fa-tumblr-square {
     color: $tumblr-color;
--- a/assets/_scss/_variables.scss
+++ b/assets/_scss/_variables.scss
@@ -81,6 +81,7 @@ $pinterest-color            : #cb2027;
 $rss-color                  : #fa9b39;
 $soundcloud-color           : #ff3300;
 $stackoverflow-color        : #fe7a15;
+$speakerdeck-color          : #396;
 $tumblr-color               : #32506d;
 $twitter-color              : #55acee;
 $vimeo-color                : #1ab7ea;
--- a/assets/_scss/vendor/font-awesome/_icons.scss
+++ b/assets/_scss/vendor/font-awesome/_icons.scss
@@ -378,6 +378,7 @@
 .#{$fa-css-prefix}-youtube-play:before { content: $fa-var-youtube-play; }
 .#{$fa-css-prefix}-dropbox:before { content: $fa-var-dropbox; }
 .#{$fa-css-prefix}-stack-overflow:before { content: $fa-var-stack-overflow; }
+.#{$fa-css-prefix}-speaker-deck:before { content: $fa-var-external-link; }
 .#{$fa-css-prefix}-instagram:before { content: $fa-var-instagram; }
 .#{$fa-css-prefix}-flickr:before { content: $fa-var-flickr; }
 .#{$fa-css-prefix}-adn:before { content: $fa-var-adn; }
--- a/_config.yml
+++ b/_config.yml
@@ -81,6 +81,7 @@ author:
   linkedin         :
   pinterest        :
   soundcloud       :
+  speakerdeck      :
   stackoverflow    : # http://stackoverflow.com/users/123456/username
   steam            :
   tumblr           :
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -64,6 +64,9 @@
       {% if author.stackoverflow %}
         <li><a href="https://www.stackoverflow.com/users/{{ author.stackoverflow }}"><i class="fa fa-fw fa-stack-overflow" aria-hidden="true"></i> Stackoverflow</a></li>
       {% endif %}
+      {% if author.speakerdeck %}
+        <li><a href="https://speakerdeck.com/{{ author.speakerdeck }}"><i class="fa fa-fw fa-speaker-deck" aria-hidden="true"></i> Speaker Deck</a></li>
+      {% endif %}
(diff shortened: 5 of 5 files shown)


assets/_scss/_utilities.scss
```diff
@@ -258,6 +258,10 @@ body:hover .visually-hidden button {
     color: $stackoverflow-color;
   }
 
+  .fa-speaker-deck {
+    #color: $speakerdeck-color;
```
jhabdas: Why is there a `#` here? That can't work, can it?

assets/_scss/_variables.scss
```diff
@@ -81,6 +81,7 @@ $pinterest-color            : #cb2027;
 $rss-color                  : #fa9b39;
 $soundcloud-color           : #ff3300;
 $stackoverflow-color        : #fe7a15;
+$speakerdeck-color          : #396;
```
jhabdas: Stylistically this should be kept as full hex.

assets/_scss/_utilities.scss
glennsarti: Typo.  I'll fix that

assets/_scss/vendor/font-awesome/_icons.scss
```diff
@@ -378,6 +378,7 @@
 .#{$fa-css-prefix}-youtube-play:before { content: $fa-var-youtube-play; }
 .#{$fa-css-prefix}-dropbox:before { content: $fa-var-dropbox; }
 .#{$fa-css-prefix}-stack-overflow:before { content: $fa-var-stack-overflow; }
+.#{$fa-css-prefix}-speaker-deck:before { content: $fa-var-external-link; }
```
jhabdas: This is clever, but hacks the library and shouldn't be added to the theme as it it'll break the moment users attempt to upgrade their version of font awesome.

assets/_scss/_variables.scss
glennsarti: Sure.  I'll update that.

jhabdas:
Thanks for providing this example. I noticed Speaker Deck hasn't been added to FA but voting is taking place here: https://github.com/FortAwesome/Font-Awesome/issues/1688.

glennsarti:
@jhabdas Yeah I did see that voting.  I also saw that the issue was raised in 2013.  And given there's over 3,500 open issues I didn't see this particular icon getting much traction in the near future.  Instead I figured on using the external link icon to provide the same functionality.

assets/_scss/vendor/font-awesome/_icons.scss
glennsarti: Ahh of course.  I should've realised that was in  a "vendored" library.  I'll revisit the way I'm doing this without hacking it in.

mmistakes:
Thanks for the PR. I'll keep this open in the off chance others are looking for a way to add supprort. I tend to shy away merging in these types of PRs as I have no interest in adding ever social network under the sun to the author sidebar. 

I'm sticking with the popular requests and the fact that Font Awesome doesn't even have an icon for Speaker Deck doesn't help its cause much :stuck_out_tongue:

glennsarti:
@mmistakes Yeah, I agree that it would be difficult to manage the links for EVERY network.  Is there perhaps a better way to display arbitrary external links in the profile;

e.g.
In a _config.yml
```
...
author:
  externallinks:
    speakerdeck: "https://......."
    someothersite: "http://...."
...
```

And then just render it as list of links in the profile?
```
<icon> Github
<icon> Twitter
<icon> speakerdeck           <----- The name of the link comes from the YML
<icon> someothersite
```

mmistakes:
Well the easiest way is hacking up `author-profile.html` just as you have. If someone wants to include other links or customize the order or any other particulars that's the best way. The theme was meant to be customized after all...

Another option is I bake in something like `_includes/custom-author-links.html` that forkers can add whatever they want without much fear of diverging from `master` if they wish to pull in updates later. I did this with the `head` and `footer` includes.

There's also support for [custom sidebars](https://mmistakes.github.io/minimal-mistakes/layout-sidebar-custom/) that could possibly be augmented to drive some of this functionality. I use them in conjunction with a custom nav lists to build the documentation site's sidebar links.

vg:
If the intention is to use a generic icon like fa-external-link for Speaker Deck, why not have a generic option itself and not limit it to speaker deck. Such that the generic option can be pointed to any url as required.

mmistakes:
@vg My intention is not to merge in these sorts of social network link adds and rather leave it to the user to customize the `author-profile.html` include. 

Editing your copy of the author sidebar affords you way more flexibility to rearrange the order and add/remove links anyway you want, rather than rely on whatever defaults I baked into the theme. 

They're there to get you started and cover the most common uses.

vg:
@mmistakes Understood. Makes sense.

glennsarti:
@mmistakes Would / Do you need assistance with https://github.com/mmistakes/minimal-mistakes/issues/362 ?

I'm happy to close this PR with a reference to that issue.

vg:
@mmistakes No, I have figured out how to customise the theme as per my needs. But I would like to see a write up on how it can be done **nicely** in the Quick Start Guide. Every change I make leads to edit conflicts which I have to resolve manually to update. That seems to be the only major problem now. Hope Jekyll 4 will improve that.

mmistakes:
@vg Once gemified Jekyll themes matures I'll be able to rebuild the theme that way. Updating and installing will be a 1,000x times easier. Plus you'll be able to override any of the layouts, includes, styles, etc. and not worry about messing with the core theme.

But that seems a bit off so for now the best you can hope for is dealing with a few merge conflicts here and there if you decide to pull in updates.

mmistakes:
@glennsarti I think I have it covered. Now #324 on the other hand... if you can help there that's one I could use some help sorting out. If it's even possible.

glennsarti:
Unfortunately I won't be much use on that one.  It's waaaaay beyond what I know.

PR #308, opened 2016-05-06
Add option for pattern-based hero headers.

tobie:
Setting the header.overlay_pattern option to true makes
the pattern repeat itself. Plays well with filters, too.

patterns


tobie:
The other option is to make `overlay_pattern` take the URL of the image.

tobie:
Hey, just curious if you're interested in this at all (either using a boolean `overlay_pattern` or one that returns a URL) and if you'd wan't me to rebase so this can be merged. Else I'll just close this.

mmistakes:
Was hesitant to include it as I don't want to over stuff everything with so many options. You can leave it open and see if anyone chimes in if this sort of thing is useful.

vg:
@tobie can you provide an unminified version of assets/css/main.css
I am not able to do a diff to see the changes

tobie:
@vg that file's generated from assets/_scss/_page.scss.

vlsi:
Just in case: I would like to have just "linear gradient" background, so regarding

> if anyone chimes in if this sort of thing is useful

I would say enabling "pattern backgound" and similar features would be valuable.

PR #316, opened 2016-05-13
Translate index page title

flohw:
I recently succeed to install and build my site with this theme (I got some issues with osx…). I have noticed that the page title of the index page was not translated.
Let me know if I can make other pull request like this or if it's preferable to "hard translate" in template inside of my project/repository.
Thanks

PS : I also have noticed a "no new line at end of file" diff but I don't know how to fix that. I am with osx and used vim to edit. I hope it's not a problem for you.


mmistakes:
Thanks for the PR @flohw

PR #325, opened 2016-05-26
Only include analytics on production environment

flohw:


mmistakes:
I'm on the fence about putting something like this into the theme.

I thought about doing exactly this with GA and the Disqus comments but was concerned that someone who doesn't host with GitHub Pages (where the `production` environment variable is set automatically) wouldn't remember or know to `jekyll JEKYLL_ENV=production build`.

And I wanted to avoid support issues that I would almost certainly receive saying "why doesn't my Google Analytics work." :neutral_face: 

It's hard enough to get someone to set `url` in `_config.yml` that I'm not those who us alternative hosting would even read my documentation to set an environment variable as part of their deployment process.

flohw:
😭

jhabdas:
**This can be closed as the backing repository has been deleted.**

Analytics can be enabled/disabled for using the `--config` flag as described in [How I'm using Jekyll in 2016](https://mademistakes.com/articles/using-jekyll-2016/). I for one would prefer not to have any magic going on behind the scenes and would lobby for always using config overrides.

PR #330, opened 2016-05-30
iOS navigation with one tap closes #315

### LINKED ISSUE #315: Two clicks required to select items on iOS devices
Your Jekyll theme is amazing!  Thank you.

Sorry if this has been addressed before.  I have to click twice on menu items and links on IOS devices.

Seems like this autocomplete bug? [https://bugs.jqueryui.com/ticket/10544](url) 

Some ideas here as well. [http://stackoverflow.com/questions/25286685/autocomplete-requires-you-to-click-twice-in-ios-after-update-to-1-11-0](url)

jhabdas:
does what is says on the box. tested on iPhone 6 via `ngrok` after running `npm run build:css`.

### DIFF
--- a/assets/_scss/_navigation.scss
+++ b/assets/_scss/_navigation.scss
@@ -234,8 +234,8 @@
         transition: $global-transition;
       }
 
-      &:hover {
-        &:before {
+      @include breakpoint($medium) {
+        &:hover:before {
           width: 100%;
         }
       }


assets/_scss/_navigation.scss
```diff
@@ -422,4 +422,4 @@
       display: block;
     }
   }
-}
\ No newline at end of file
+}
```
jhabdas: Just saw `insert_final_newline = false` in your `.editorconfig` file and had to do a double take on my editor. But it's enabled. YOLO

mmistakes:
I think there's a way to do this and preserve the hover effect... Even on touch enabled devices. On my personal site mademistakes.com I do pretty much the same effect on the main navigation links and there's no double tap on iOS. Just need to adjust the CSS some to match.

jhabdas:
Look at that. And it works on WebKit Mobile. Sweet. Will close this for now since we know there are other ways to resolve

PR #335, opened 2016-06-02
javascript to embed comments from discourse forum

rriemann:
The comment system can be changed to embed topics from the Discourse
forum software: http://discourse.org/

More details: https://meta.discourse.org/t/embedding-discourse-comments-via-javascript/31963

### DIFF
--- a/_includes/comments-providers/discourse.html
+++ b/_includes/comments-providers/discourse.html
@@ -1,0 +1,13 @@
+{% if site.comments.discourse.server %}
+{% capture canonical %}{{ site.url }}{% if site.permalink contains '.html' %}{{ page.url }}{% else %}{{ page.url | remove:'index.html' | strip_slash }}{% endif %}{% endcapture %}
+<script type="text/javascript">
+    DiscourseEmbed = { discourseUrl: '//{{ site.comments.discourse.server }}/',
+                       discourseEmbedUrl: '{{ canonical }}' };
+   (function () {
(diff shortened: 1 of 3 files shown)


_includes/comments-providers/discourse.html
```diff
@@ -0,0 +1,13 @@
+{% if site.comments.discourse.server %}
+{% capture canonical %}{{ site.url }}{% if site.permalink contains '.html' %}{{ page.url }}{% else %}{{ page.url | remove:'index.html' | strip_slash }}{% endif %}{% endcapture %}
```
mmistakes: Might want to add `{{ site.baseurl }}` after `{{ site.url }}` as the theme supports that and if defined in `_config.yml` will properly assign the wrong page url to `canonical`.

rriemann:
Ok, I reuse the base_path snippet.

mmistakes:
@rriemann I've never setup Discourse before. Are the resoruces on https://meta.discourse.org/t/embedding-discourse-comments-via-javascript/31963 all that someone needs to get up and running? I want to be sure I include all the relevant info in the theme's docs because I know I'll get support issues about it eventually :wink:.

rriemann:
Yes, it should be fine. The Discourse meta forum is a great place to figure out 
problems and if there is a problem at some point, the provided link is not 
only the right place to find some advice, but also to ask questions.

PR #353, opened 2016-06-11
Make the location section of the author sidebar not have a hand cursor

amweiss:
I didn't see a contributing section section, so hopefully this and my other PR work for you.


amweiss:
Just noticed it had my configs, I'll fix that.

PR #360, opened 2016-06-14
Adding 500px as social profile

basnijholt:


mmistakes:
Closing to keep the theme lean and avoid configuration bloat. The defaults cover the most popular networks, anything else I envisioned being customized by the user.

Issue #372, opened 2016-06-22
Clicking Follow Button and resizing removes links.


ZenHarbinger:
- [ ] This is a question about using the theme.
- [x] I believe this to be a bug with the theme --- not Jekyll, GitHub Pages or one of the bundled plugins.
- [ ] This is a feature request.
- [x] I have updated all gems with `bundle update`.
- [x] I have tested locally with `bundle exec jekyll build`.

---

## Environment informations

- **Minimal Mistakes version:** master
- **`github-pages` or `jekyll` gem version:** 
- **Operating system:** Ubuntu Linux 64-bit

---

## Expected behavior

When the Follow button is clicked to display the list of options and then the window is resized to allow the removal of the Follow button, all of the links should re-appear.

## Steps to reproduce the behavior

1. Resize page to make the Follow button appear
2. Click on the Follow button
3. Click again to make the links go away.
4. Resize the page to remove the Follow button
5. No links

![screenshot from 2016-06-22 18-09-41](https://cloud.githubusercontent.com/assets/1130510/16285507/84fd606c-38a4-11e6-8d58-5dcc2ac6ffb6.png)

PR #373, opened 2016-06-22
Fixed disappearing menu.

ZenHarbinger:
Fix for issue 372

https://github.com/mmistakes/minimal-mistakes/issues/372

Force .author_urls to show() when width is wide enough.
Force .author_urls to hide() when not.


ZenHarbinger:
Switch from using window width to button visibility for if the author urls are shown.

ZenHarbinger:
Sorry for the redundant pulls.
Cleaned up _main.js
Added code straight to main.min.js

mmistakes:
Thanks. I merged this in before you updated the pull so if you wouldn't mind resubmitting it since I updated some JS dependencies that are likely missing in your build of `main.min.js`. If you pull from `master` you should get them and then I can merge in the updated PR.

PR #374, opened 2016-06-23
Use of button visibility instead of window width

ZenHarbinger:
Display author urls using button viz instead of window width.


_includes/author-profile.html
```diff
@@ -1,6 +1,6 @@
 {% include base_path %}
 
-{% if page.author %}
+{% if page.author and site.data.authors[page.author] %}
```
mmistakes: What was the thinking behind changing this line? What happens if someone doesn't create an `authors.yml` data file?

ZenHarbinger: That is exactly my thinking behind this line. I don't have an authors.yml file, so it was failing.  I added the check so that it would default back to the _config.yml settings.

ZenHarbinger:
Have created 2 separate PRs.  Should be able to merge at your leisure independently.

PR #375, opened 2016-06-24
Updated check for button visibility instead of window width.

ZenHarbinger:
PR for check on button visibility instead of width.
Allows user the ability to change the width at which the widget changes.


mmistakes:
Thanks again for this and the other PR @ZenHarbinger

ZenHarbinger:
No problem!

Issue #378, opened 2016-06-25
Different feed providers


gabor-meszaros:
- [ ] This is a question about using the theme.
- [ ] I believe this to be a bug with the theme --- not Jekyll, GitHub Pages or one of the bundled plugins.
- [x] This is a feature request.
- [x] I have updated all gems with `bundle update`.
- [x] I have tested locally with `bundle exec jekyll build`.

---
Please make it possible to choose from different feed providers (e.g. feedburner) just as you can choose from different comment providers.

PR #379, opened 2016-06-25
Add different feed providers

gabor-meszaros:
Solution for #378 . Could you please review it? I am quite new in Jekyll and Liquid. :) Thank you!


mmistakes:
I'm onboard with this enhancement. One small suggestion. What of instead of doing something similar to the comments provider customization it's something much simpler:

1. By default use `feed.xml` that builds with the `jekyll-feed` plugin.
2. All the default `feed.xml` to be overriden by specifying a path to an external one eg: http://feeds.feedburner.com/whatever or anywhere else.

Removing the provider stuff makes it easier to use any other service.

Thinking `_config.yml` would look something like

```yaml
feed:
  path: # blank (default) uses feed.xml
```

And `_includes/footer.html`

```html
  <li><a href="{% if site.feed.path %}{{ site.feed.path }}{% else %}{{ base_path }}/feed.xml{% endif %}"><i class="fa fa-fw fa-rss-square" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].feed_label }}</a></li>
```

gabor-meszaros:
Thank you for the review! It makes sense and indeed less complicated than my original solution.

I did the modifications. Is it anything else that I can do to improve it?

_config.yml
```diff
@@ -29,9 +29,7 @@ comments:
     num_posts            : # 5 (default)
     colorscheme          : # "light" (default), "dark"
 feed:
```
mmistakes: Testing this I hit an error. Maybe change `feed` to `atom_feed` since I believe the previous name was conflicting with the `jekyll-feed` plugin.

Also if you could squash all the commits into 1 with rebase that would make things cleaner when I merge the pull request in.

Thanks!

_includes/footer.html
```diff
@@ -17,8 +17,8 @@
     {% if site.author.bitbucket %}
       <li><a href="http://bitbucket.org/{{ site.author.bitbucket }}"><i class="fa fa-fw fa-bitbucket" aria-hidden="true"></i> Bitbucket</a></li>
     {% endif %}
-    <li><a href="{{ base_path }}/feed.xml"><i class="fa fa-fw fa-rss-square" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].feed_label }}</a></li>
+    <li><a href="{% if site.feed.path %}{{ site.feed.path }}{% else %}{{ base_path }}/feed.xml{% endif %}"><i class="fa fa-fw fa-rss-square" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].feed_label }}</a></li>
```
mmistakes: Change variable name from `feed` to `atom_feed` to avoid conflicting with `jekyll-feed` plugin.

gabor-meszaros:
Thank you for the suggestions! I am sorry for the quite late answer...

I noticed that some commits were added to the repository since my last commit, I am not sure if I can do a rebase in this situation. If it is possible, could you please give me some hints about how to do it?

I do not want to hack around if there is a better solution to this, but I can fork your repo again, do the modification, and create a new PR if that makes the git log more readable.

What is your opinion? Thanks!

mmistakes:
Rebasing and flattening everything down into a single commit would probably be cleaner. But I'm not the best at Git and haven't had much luck rebasing myself.

If you don't mind, probably easier to just refork and open a new PR.

gabor-meszaros:
Sure, I will do it right now. Thanks for your patience!

gabor-meszaros:
I added the new PR. Could you please review it whether it is OK? Thanks.

PR #398, opened 2016-07-14
Fix to sidebar

ZenHarbinger:
Added check to see if the author button is present.
If it is not, then the width is used (as it was).
If it is, then it uses the button visibility.

It was working for me on your page (no more overlay).


mmistakes:
Thanks for jumping in and fixing this so quickly :smile:

ZenHarbinger:
No problem, would have done it a bit quicker except for kids and getting a run in while it's not murderous outside.
Let me know if there are any other problems and I'll hit them too.

mmistakes:
I know how all that goes, especially the kids part.

PR #400, opened 2016-07-17
Display post date for each post at archive

jenglamlow:
Just a personal preference of displaying post date on the archive page, instead of navigate into the post to search for the post date


mmistakes:
The archive single include can be used in collections and not just posts. So there is no guarantee a `post.date` will exist in every situation.

jenglamlow:
Is there a plan to have setting to display the post.date for archive?

mmistakes:
Not currently. I didn't include it as a default because it didn't add much value and is quite easy to add (as you've done in your PR) if someone wants it.

Theme is a starting point, meant to hack on. I'm trying to avoid adding every possible option. For every person who will want the dates to show, there will be someone who wants them hidden. Making them configurable adds more complexity to the theme.

jenglamlow:
Ok, keep it that way then. You may close this pull request. 

Thanks for your nice theme. Really appreciate it

PR #403, opened 2016-07-24
Update ui-text.yml to add "tr" locale

ekremkaraca:
Turkish locale values...


mmistakes:
Thanks for the PR @ekremkaraca

PR #406, opened 2016-07-26
Add external atom feed URL config possibility

gabor-meszaros:
Solution for #378 based on the discussions in #379 .


mmistakes:
Thanks for resubmitting the PR. When testing I noticed it was missing a feed link in [`head.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/head.html#L7) so I took care of that when I merged this in.

Thanks again.

PR #411, opened 2016-07-31
 Adding Portugese in the ui-text

lipemorais:
I'm from Brasil and added the Portuguese to ui-text file


mmistakes:
Thanks for the PR @lipemorais. This will be merged in with the next release.

Issue #462, opened 2016-08-18
Underscores & Hyphens in the og:locale meta tag


carpiediem:
- [ ] This is a question about using the theme.
- [x] I believe this to be a bug with the theme --- not Jekyll, GitHub Pages or one of the bundled plugins.
- [ ] This is a feature request.
- [x ] I have updated all gems with `bundle update`.
- [x] I have tested locally with `bundle exec jekyll build`.

---

## Environment informations

Running on github.io

---

## Expected behavior

og:locale meta tag should be processed

## Steps to reproduce the behavior

If you test a page in https://developers.facebook.com/tools/debug/sharing/, it shows that the og:locale tag is being rejected.  Facebook is expecting a value with an underscore (e.g. "en_US") but the default configuration of minimal-mistakes uses a hyphen (e.g. "en-US").

Simply modifying the _config.yml file isn't an option, since ui-text.yml is written for hyphens.  I think those are the only two files that would need to be edited, though.

mmistakes:
Should be an easy fix adding a `replace` filter and changing the default to `en_US` to this line in `_includes/seo.html`

```liquid
<meta property="og:locale" content="{{ site.locale | replace: "-", "_" | default: "en_US" }}">
```

PR #465, opened 2016-08-19
formats og:locale tag with underscores

carpiediem:
Addresses issue #462 

See http://fbdevwiki.com/wiki/Locales


mmistakes:
Thanks for fixing this.

PR #466, opened 2016-08-19
Additional options for images to use in social share meta tags

carpiediem:
Use page.header.teaser instead of site.og_image in twitter:image and  og:image meta tags


mmistakes:
Thanks for the PR!

PR #472, opened 2016-08-23
Add french translations

flohw:
Sorry for delay, was on holidays :)
I think translations are OK but I was not able to run `bundle exec jekyll s` after pulling changes on your repository. Got this message :
```
 ERROR: YOUR SITE COULD NOT BE BUILT:
                    ------------------------------------
                    Invalid date '<%= Time.now.strftime('%Y-%m-%d %H:%M:%S %z') %>': Document 'vendor/ruby/2.3.0/gems/jekyll-3.1.6/lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb' does not have a valid date in the YAML front matter.
```

I havn't enough time to search now. I have run `bundler install`

### DIFF
--- a/_data/ui-text.yml
+++ b/_data/ui-text.yml
@@ -120,17 +120,18 @@ fr: &DEFAULT_FR
   email_label                : "Email"
   recent_posts               : "Posts récents"
   undefined_wpm              : "Le paramètre words_per_minute n'est pas défini dans _config.yml"
-  comment_form_info          :
-  comment_form_comment_label :
-  comment_form_md_info       :
-  comment_form_name_label    :
-  comment_form_email_label   :
-  comment_form_website_label :
-  comment_btn_submit         :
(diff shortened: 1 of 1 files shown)


mmistakes:
Thanks.

I don't think that error is theme related. Jekyll has a problem with `vendor/ruby/2.3.0/gems/jekyll-3.1.6/lib/site_template/_posts/0000-00-00-welcome-to-jekyll.markdown.erb` which is in your system, not the theme.

Probably have to throw `vendor` into `exclude` in _config.yml to ignore that file. But doing that may block the `/assets/js/vendor/`, which will probably need to be put in `include` so Jekyll processes those files.

flohw:
Fixed, I used the default configuration. Then some documentation to read with staticman and update ui-text.
You can merge that :smile:

mmistakes:
Merged. Thanks!

PR #474, opened 2016-08-24
Fixed error link "posts" in breadcrumbs "home/posts/page". 

Rean1030:
When click "posts" link in breadcrumbs path "home/posts/page". It will
turn to an error path like:
"https://username.github.ioposts" not as a right one
"https://username.github.io/posts".


Rean1030:
Oh, So sorry for my mistake. I just want to commit the last one 😫......

Rean1030:
I committed it in the develop branch instead of here . By the way added simplified Chinese in UI-text.

PR #482, opened 2016-08-26
Add Chinese in UI-text & Fixed error link "posts".

Rean1030:
1.Fixed error link "posts" in breadcrumbs "home/posts/page". 

2.Add simplified Chinese in UI-text.


mmistakes:
Could you breakout try the Chinese text translation commit as its own pull request?

Could you clarify what the breadcrumb commits are all about? Not exactly sure what that's suppose to fix. Looking at it you removed some conditionals that will give collection pages a problem if breadcrumbs are enabled.

Rean1030:
Sorry for these, I will  look over carefully next time.

PR #487, opened 2016-08-28
clarified comment for author.stackoverflow value in _config.yml

rschaerer:
# Initial Situation

* The author.stackoverflow variable was commentet with `# http://stackoverflow.com/users/123456/username`, which led me to the conclusion that I should add the whole url as value.
* The stackoverflow combination of numeric id / username isn't self-explanatory.

# Proposed Change

Change the comment to explicit state the requested value `"123456/username"` and append the whole demo url to show where the needed value can be found.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -97,7 +97,7 @@ author:
   linkedin         :
   pinterest        :
   soundcloud       :
-  stackoverflow    : # http://stackoverflow.com/users/123456/username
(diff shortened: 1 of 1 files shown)


mmistakes:
Thanks. Merged into `develop` and will push with the next theme release.

ref: https://github.com/mmistakes/minimal-mistakes/commit/d3ac1123da4a2e6b2b75d8de0d46a36c57bda576

PR #488, opened 2016-08-29
Add list of available translations

rschaerer:
Since there are several translations in the master branch, they should also be featured in the docs.

For further additions I styled them as a list in alphabetical order.

The native language description (e.g. Français) might make it easier to find this page for search queries in the native language. But I'm not sure if it is worth the effort...


mmistakes:
Excellent. Thanks for the added documentation!

PR #496, opened 2016-09-01
get Follow-Button label from ui-text and default to "Follow"

rschaerer:
The Follow button Label wasn't localized with the ui-text value, now it is with default to "Follow".

The colon from the end of the ui-text value of `follow_label` is being removed by the liquid filter `remove: ":"`


mmistakes:
Thanks for the PR!

PR #514, opened 2016-09-09
Completed Italian translation

nmiglio:
Thanks for this nice template. Here is my little contribution.


mmistakes:
Fantastic. Thank you @nmiglio!

PR #552, opened 2016-09-23
sidebar image error

arjitb:
This code was already present in the gh-pages(as demo'd in https://mmistakes.github.io/minimal-mistakes/portfolio/foo-bar-website/), but after some looking around I found that the code that is being used by other users from the master repo has a bug while the gh-pages code for this file is correct. 
The bug is that when viewed on a mobile device or a smaller screen width for that matter, the sidebar doesn't stop the animation, so the image overlaps the text leaving the text not viewable.


mmistakes:
Hard to tell what was fixed. Can you be a little more specific on what the issue is this PR addresses? 

If you could also make changes to the [source JS](https://github.com/mmistakes/minimal-mistakes/blob/master/assets/js/_main.js) files as well as the compiled `main.min.js` file it would appreciated.

PR #558, opened 2016-09-27
Update seo.html for meta description

justinrummel:
Adding standard meta description immediately after seo_description has been defined


mmistakes:
Thanks @justinrummel

Issue #560, opened 2016-09-28
Easier image insertion in posts


fa-ribeiro:
Based on [eduardo bouças blog post](https://eduardoboucas.com/blog/2014/12/07/including-and-managing-images-in-jekyll.html), I created an `image` tag to simplify image insertion on posts (with and without captions). It presumes all images are placed under `asset/images` and all I have to do is to specify image path from that point. The syntax is something like this:

    {% include image name="2016/07/MyPostImage.png" caption="The optional image caption" %}

The include file `image` looks like this:

   {% if include.caption %}
     <figure>
       <img src="{{ site.url }}{{ site.baseurl }}/assets/images/{{ include.name }}" {% if include.alt %} alt="{{ include.alt }}" {% endif %} {% if include.width %} width="{{ include.width }}" {% endif %}/>
       <figcaption>{{ include.caption }}</figcaption>
     </figure>
   {% else %}
     <img src="{{ site.url }}{{ site.baseurl }}/assets/images/{{ include.name }}" {% if include.alt %} alt="{{ include.alt }}" {% endif %} {% if include.width %} width="{{ include.width }}" {% endif %}/>
   {% endif %}

I'me wondering if it would be interesting to add this feature to the theme. if yes, I can submit a PR.

What do you think?

mmistakes:
Could be a nice compliment to the [gallery helper](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/gallery). If you want to work on a PR that would be cool. I'd suggest trying to use some of the same variable names to keep things consistent.

PR #572, opened 2016-10-07
add image helper for easier image insertion in posts

fa-ribeiro:
This PR adds an 'image' helper for easier insertion of image in posts (#560)

It handles both local images (it assumes images are under `assets/images`) and urls.

in order to add an image to a post, just use the following syntax:

`{% include image image_path="2016/07/MyPostImage.png" caption="The optional image caption" %}`

or

`{% include image image_path="https://example.com/images/MyPostImage.png" caption="The optional image caption" alt="The optional alt text" %}`

Parameters are:
- `image_path` (mandatory): the image path. It can be an url or a local path. In the former case, it assumes images are under `assets/images`, so you only need to specify the path from there.
- `caption` (optional): the image caption.
- `alt` (optional): the alt text.

What you think?

### DIFF
--- a/_includes/image
+++ b/_includes/image
@@ -3,7 +3,7 @@
     {% if include.image_path contains "://" %}
       "{{ include.image_path }}"
     {% else %}
-      "{{ include.image_path | prepend: "/assets/images/" | absolute_url }}"
+      "{{ include.image_path | absolute_url }}"
     {% endif %}
     alt="{% if include.alt %}{{ include.alt }}{% endif %}">
   {% if include.caption %}
--- a/_includes/image
+++ b/_includes/image
@@ -1,11 +1,9 @@
-{% include base_path %}
-
 <figure class="{{ include.class }}">
   <img src=
     {% if include.image_path contains "://" %}
       "{{ include.image_path }}"
     {% else %}
-      "{{ include.image_path | prepend: "/assets/images/" | prepend: base_path }}"
+      "{{ include.image_path | prepend: "/assets/images/" | prepend: absolute_url }}"
     {% endif %}
     alt="{% if include.alt %}{{ include.alt }}{% endif %}">
   {% if include.caption %}
--- a/_includes/image
+++ b/_includes/image
@@ -1,0 +1,14 @@
+{% include base_path %}
+
+<figure class="{{ include.class }}">
+  <img src=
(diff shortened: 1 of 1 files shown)


mmistakes:
Looks good and fits in well with the `gallery` helper. I'll see about merging this in when I do the next release of the theme. Currently working on a few things to make the switch over to a gemified version of the theme and I'd like to include this with that release.

_includes/image
```diff
@@ -0,0 +1,14 @@
+{% include base_path %}
```
mmistakes: I've deprecated the `base_path` include as Jekyll provides filters to do this now.

_includes/image
```diff
@@ -0,0 +1,14 @@
...
+
+<figure class="{{ include.class }}">
+  <img src=
+    {% if include.image_path contains "://" %}
+      "{{ include.image_path }}"
+    {% else %}
+      "{{ include.image_path | prepend: "/assets/images/" | prepend: base_path }}"
```
mmistakes: can replace `prepend: base_path` with `absolute_url`

fa-ribeiro:
done :-)

fa-ribeiro:
done :-)

_includes/image
```diff
@@ -0,0 +1,12 @@
+<figure class="{{ include.class }}">
+  <img src=
+    {% if include.image_path contains "://" %}
+      "{{ include.image_path }}"
+    {% else %}
+      "{{ include.image_path | prepend: "/assets/images/" | absolute_url }}"
```
mmistakes: One more minor thing. With the latest MM I moved away from assuming images would be placed in `/assets/images/` and instead encourage adding the full path. Adds more flexibility to where they can be placed for the user.

If you could remove `prepend: "/assets/images/"` it fall in line with the [other helpers](https://mmistakes.github.io/minimal-mistakes/docs/helpers/) and I can merge it in.

mmistakes:
Looks like Jekyll core may get a native filter to do something similar. https://github.com/jekyll/jekyll/pull/5560

mmistakes:
I made a small change and renamed the helper from `image` to `figure` to be more semantic. The PR has been merged c4252b09afaf5794d52248b34352b2c2e6c10336 and will show up in the next theme release.

Thanks for your work on this.

PR #576, opened 2016-10-08
Repair aggressive box-shadow in navigation.

maciejczyzewski:
Before:

screenshot 2016-10-08 22 47 14

After:

screenshot 2016-10-08 22 47 03


mmistakes:
Thanks for the PR!

PR #586, opened 2016-10-14
disqus: Use canonical URL and whitespace clean-up

kylemanna:
Fix an issue that would cause Disqus to server different comments for every possible variation of URL (defaulted to `window.location.href`) including things such as GET parameters, Google translate, localhost development servers and more.

Clean-up white space while we're at it.

See individual commits for more detail.

### DIFF
--- a/_includes/comments-providers/disqus.html
+++ b/_includes/comments-providers/disqus.html
@@ -1,27 +1,25 @@
 {% if site.comments.disqus.shortname %}
 {% capture canonical %}{{ base_path }}{% if site.permalink contains '.html' %}{{ page.url }}{% else %}{{ page.url | remove:'index.html' | strip_slash }}{% endif %}{% endcapture %}
   <script type="text/javascript">
-  	/* * * CONFIGURATION VARIABLES: EDIT BEFORE PASTING INTO YOUR WEBPAGE * * */
-  	var disqus_shortname = '{{ site.comments.disqus.shortname }}';
+    var disqus_shortname = '{{ site.comments.disqus.shortname }}';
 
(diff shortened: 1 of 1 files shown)


kylemanna:
Hello, is there anything I can do to help this get merged?

mmistakes:
@kylemanna Sorry haven't had much time to look at this. I think it might be a breaking change that I'd likely hold off on till I roll out the Gemified version of the theme.

Reason I think it may be a breaker is due to adding the canonical URL instead of letting Disqus determine the page URL itself. I haven't tested it, but if those values are different then it's going to break things for users.

Meaning if they have existing comments for a page and the canonical URL is different their comments will disappear, unless they want to use Disqus' migration tool.

I've been bit by this one a few times when changing things related to permalinks so just want to make sure it doesn't cause a ruckus.

kylemanna:
Hi @mmistakes I think you're right.  Some percentage of sites with correct canonical URLs will work better and sites with invalid canonical URLs would require the dev to migrate the Disqus links and correct their canonical URL.

Welcome to close this pull request if you deem it necessary. 

That said, it would be nice to work in to your great project at some point as it would have helped keep my Disqus threads more sane if I had it enabled from the get go.

kylemanna:
Perhaps a compromise could be to create a flag that is disabled if unset, and initialize the `_config.yml` to  enable it.

That way people who blindly merge this would have the previous functionality, and new adopters would start off on the right foot.

I can put this together if this is something you'd like to merge.

mmistakes:
That sounds like a good way to approach it. As long as it doesn't overly complicate things making it harder to maintain then I say go for it.

PR #595, opened 2016-10-16
Add an option to allow modders to change date formatting across the layout

vndmtrx:
With this change, we can better control when date is used on the layout, avoiding the necessity to edit the layout files into the desired format schema.


mmistakes:
What are the read time changes about? Doesn't seem related to the date time format config.

vndmtrx:
Ohh, sorry, I thought I had done it in another branch. Fixed.

PR #611, opened 2016-10-24
Fix/kaggle images

ivyleavedtoadflax:
Fix images in kaggle post


ivyleavedtoadflax:
sorry my bad - wrong destination!

Issue #324, opened 2016-05-26
Replace custom seo.html with Jekyll SEO Tag plugin


ddd-workshop:
Recently announced by GitHub: https://github.com/blog/2162-better-discoverability-for-github-pages-sites

mmistakes:
Sort of. I originally planned to use the Jekyll SEO Tag plugin that comes along with GitHub Pages but it didn't fully support everything the theme does (eg: Twitter Summary cards with large images).

So I took what the plugin did and baked it into the theme so I could enhance and build off of it.

If for whatever reason you'd rather use the native plugin instead of my customizations you can rip out `{% include seo.html %}` from `/_includes/head.html` and replace with `{{ seo }}` and you're good to go.

mmistakes:
**Followup.** The Jekyll SEO Tag plugin has made a lot of nice updates and now supports everything I was custom rolling into the theme and would love to swap it in.

Only thing is it would be a bit more work if you want to include feature images as meta data for Twitter Cards and Open Graph. The variable names are different so you'd have to manually specify them and edit any old content.

| Jekyll SEO Tag | MM theme |
| ------------------- | -------------- |
| `page.image` | `page.header.image` |
| `site.logo` | `site.og_image` |

I tried using some Liquid `assign` tags to override variables and placed that before `{% seo %}` in `head.html` but it didn't seem to work.

```yaml
{% if page.header.image contains "://" %}
  {% assign page.image = page.header.image %}
{% else %}
  {% assign page.image = page.header.image | prepend: "/images/" | prepend: base_path %}
{% endif %}

{% if site.og_image contains "://" %}
  {% assign site.logo = site.og_image %}
{% else %}
  {% assign site.logo = site.og_image | prepend: "/images/" | prepend: base_path %}
{% endif %}

{% seo %}
```

If anyone has any ideas of how to pull in images to be used with `{{ seo }}` without having to change variable names in the theme or edit YAML Front Matter in posts/pages/collections that would be great.

mmistakes:
Workaround. Use YAML anchors to avoid duplicating values.

Example:

```yaml
header:
  image: &image /assets/images/filename.jp
image: *image
```

or

```yaml
header:
  image: &image /assets/images/filename.jp
image: 
  path: *image
```

jhabdas:
@mmistakes need a hand with this? It'd be great to get this into the theme. Related discussion occurring here: https://github.com/jekyll/jekyll-seo-tag/issues/159

[...]

PR #665, opened 2016-11-21
Upgrades Schema Metadata

aloisklink:
### This pull request upgrades Schema Metadata to better allow Google Search to parse data in the website.

Firstly in 23af38b, it changes the [`_includes/author-profile.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/author-profile.html) file so that the author-profile has the Schema `itemprop="author"`. This does not affect anything, except when a page uses `layout: single`. In this case, the page's `CreativeWork` Schema object will list the author-profile as the `author` of the `CreativeWork`.

Secondly in b9f07ac, the `seo_author` field in [`_includes/seo.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/seo.html) becomes `site.author` if none of the other fields are filled. 
Before, the last value it would try to take was `site.author[0]`. However, in the default [`_config.yml`](https://github.com/mmistakes/minimal-mistakes/blob/master/_config.yml) file, `site.author` is not an array. So the code has been changed so that if `site.author[0]` is null, `site.author` will be used instead.

Thirdly in 	878100d, some Schema details of type `WebSite` were added to [`_includes/seo.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/seo.html). This should [allow Google to add the site-name to search results](https://developers.google.com/search/docs/data-types/sitename).

Finally in 	9456f0c, the Schema details of type `{% if site.social.type %}{{ site.social.type }}{% else %}Person{% endif %}` were modified, so that if `site.social.name` was not set, schema information would be set using the `seo_author` field. All of the fields added are almost exactly the same as the Schema fields in [`_includes/author-profile.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/author-profile.html), however, only [the social profiles that Google Search supports](https://developers.google.com/search/docs/data-types/social-profile-links) were added.

### DIFF
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -4,7 +4,7 @@
   {% assign author = site.author %}
 {% endif %}
 
-<div itemscope itemtype="http://schema.org/Person">
+<div itemprop="author" itemscope itemtype="http://schema.org/Person">
 
   {% if author.avatar %}
     <div class="author__avatar">
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -125,9 +125,67 @@
     {
       "@context" : "http://schema.org",
       "@type" : "{% if site.social.type %}{{ site.social.type }}{% else %}Person{% endif %}",
-      "name" : "{{ site.social.name | default: site.name }}",
-      "url" : {{ seo_url | jsonify }},
-      "sameAs" : {{ site.social.links | jsonify }}
+      
+      {% if site.social.name %}
+        "name" : {{ site.social.name}},
+        "sameAs" : {{ site.social.links | jsonify }},
+        "url" : {{ seo_url | jsonify }}
+      {% else %}
+      
+        "name" : "{{ seo_author.name }}",
+        {% if seo_author.avatar %}
+        "image" : "{{ seo_author.avatar | absolute_url }}",
+        {% endif %}      
+        {% if seo_author.email %}
+        "email" : "{{ seo_author.email }}",
+        {% endif %}      
+        {% if seo_author.location %}
+        "homeLocation" : { "@type" : "Place",
+                           "name" : "{{ seo_author.location }}"
+                         },
+        {% endif %}  
+        {% if seo_author.bio %}
+        "description" : "{{ seo_author.bio }}",
+        {% endif %}
+        
+        "sameAs " : [
+          {% capture seo_social_media %}
+            {% if seo_author.facebook %}
+              "https://www.facebook.com/{{ seo_author.facebook }}",
+            {% endif %}
+            {% if seo_author.twitter %}
+              "https://twitter.com/{{ seo_author.twitter }}",
+            {% endif %}
+            {% if seo_author.google_plus %}
+              "https://plus.google.com/+{{ seo_author.google_plus }}",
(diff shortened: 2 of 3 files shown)


mmistakes:
What's your thought on the [Jekyll SEO tag plugin](https://github.com/jekyll/jekyll-seo-tag/)? I'm mulling over pulling out all the schema stuff and using it instead. 

1. It covers a lot of what I'm already doing with the theme
2. One less thing to worry about maintaining in the code
3. Removes complexity from theme's `_includes` and DRYs up the code.
4. Is documented better than my theme with regards to SEO.

The plugin adds the same schema markup through JSON-LD which can be customized more than what I'm offering with the theme (eg. change type per page `BlogPosting`, `NewsArticle`, `Person`, `Organization`.)

For example instead of adding all the schema markup to the theme's includes as you've done with this PR, you can assign all the author social links like this with the plugin:

```yaml
social:
  name: Ben Balter
  links:
    - https://twitter.com/BenBalter
    - https://www.facebook.com/ben.balter
    - https://www.linkedin.com/in/BenBalter
    - https://plus.google.com/+BenBalter
    - https://github.com/benbalter
    - https://keybase.io/benbalter
```

It's not automatic like with your PR, but offers the same without adding complexity to the theme's code.

The one thing that's been holding me back was how it handles [OpenGraph/Twitter Card images](https://github.com/mmistakes/minimal-mistakes/issues/324). They're using `image` vs. `header.image`  to assign the path which means for those who use the existing theme they'd have to make some YAML changes to all of their posts. I don't think that's the end of the world (I've done it to my personal site with over 1,000 posts so it's possible).

aloisklink:
I think it would be a good idea to use it.

The Jekyll SEO tag plugin [works with GitHub Pages](https://pages.github.com/versions/).

And as far as I'm aware, all of the formatting in the [`_config.yml`](https://github.com/mmistakes/minimal-mistakes/blob/master/_config.yml) matches the SEO tag plugin, so there shouldn't be any need for anyone to update their `_config.yml` files. (Except for changing `og_image: ` to `logo: `, but that will be easily seen and fixed when merging) 

Regarding the `image` vs. `header.image` issue, I would have recommended that Liquid `assign` should be used to fix it, but [you've already tried that](https://github.com/mmistakes/minimal-mistakes/issues/324). However, changing those tags in the markdown files should be pretty easy. One would probably only need to run something like
`replace "    image: " "image: " -- *.md` or `sed -i -e 's/"    image: "/"image: "/g' *.md` in bash, if you don't need to use `header.image` anywhere, and using [YAML anchors ](https://github.com/mmistakes/minimal-mistakes/issues/324) should require only a simple Python script. 

The SEO tag plugin doesn't seem to do most of [Google Search's recommended markup](https://developers.google.com/search/docs/guides/intro-structured-data), but then again, even if one does have it on their page, there is no guarantee that Google Search will show it. Also, I'm guessing that people who care about Google Search's markup could either add it in manually, or contribute and add it to the Jekyll SEO tag plugin.

mmistakes:
Thanks for the feedback... pretty much echoes what I was thinking.

PR #685, opened 2016-11-29
Slight corrections/improvments with French

DEL-R:
- `comment_success_msg` : Just corrected a missing `m` in *commentaire*

- `related_label` : Added `aussi` (*as well*) in the end of the sentence to make it more "natural" French

- `less_than` : `plus petit que` wasn't right in this precise context - this form is more about the mathematical character `<` - `moins de` is one of the exact use in this particular case (talking about duration)                    
*Less than one minute* == `Moins de 1 minute`

- `follow_label` : Using `Contact` in place of `Suivez moi` make it less personal, and so, better suited for most cases (`Suivez moi` is more like `Follow me`)

________________

Anyhow, thank you very much for your hard work

### DIFF
--- a/_data/ui-text.yml
+++ b/_data/ui-text.yml
@@ -105,7 +105,7 @@ fr: &DEFAULT_FR
   menu_label                 :
   toc_label                  : "Sur cette page"
   ext_link_label             : "Lien direct"
-  less_than                  : "plus petit que"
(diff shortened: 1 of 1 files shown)


mmistakes:
Looks good to me. Thanks for the PR :+1:

flohw:
:+1:

PR #746, opened 2016-12-31
Complete Simplified Chinese Translation of Theme UI

vejuhust:


vejuhust:
Sorry. Send the wrong one.

PR #747, opened 2016-12-31
Complete Simplified Chinese Translation of Theme UI

vejuhust:


mmistakes:
Thanks for this @vejuhust

PR #757, opened 2017-01-07
Pagination now works for index.html files in subfolders.

maik:
I wanted to use the minimal-mistakes theme for a blog with a landing page. Hence, I've used the splash layout for the index.html page in the root folder and I've created a subfolder named blog containing another index.html file. The latter index.html file should show the most recent blog posts and it should also paginate them. That wasn't possible using the original paginator.html file, because it did not use the paginate_path variable.

### DIFF
--- a/_includes/paginator.html
+++ b/_includes/paginator.html
@@ -60,7 +60,7 @@
 
     {% comment %} Link next page {% endcomment %}
     {% if paginator.next_page %}
-      <li><a href="{{ paginator.next_page_path }}">{{ site.data.ui-text[site.locale].pagination_next | default: "Next" }}</a></li>
+      <li><a href="{{ site.paginate_path | replace: ':num', paginator.next_page | replace: '//', '/' | absolute_url }}">{{ site.data.ui-text[site.locale].pagination_next | default: "Next" }}</a></li></a></li>
     {% else %}
       <li><a href="#" class="disabled"><span aria-hidden="true">{{ site.data.ui-text[site.locale].pagination_next | default: "Next" }}</span></a></li>
     {% endif %}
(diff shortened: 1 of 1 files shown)


_includes/paginator.html
```diff
@@ -54,15 +55,15 @@
...
       <li><a href="#" class="disabled current">{{ paginator.page }}</a></li>
     {% else %}
-      <li><a href="{{ '/page' | absolute_url }}{{ paginator.total_pages }}/">{{ paginator.total_pages }}</a></li>
+      <li><a href="{{ site.paginate_path | replace: ':num', paginator.total_pages | replace: '//', '/' | absolute_url }}">{{ paginator.total_pages }}</a></li>
     {% endif %}
 
     {% comment %} Link next page {% endcomment %}
     {% if paginator.next_page %}
-      <li><a href="{{ '/page' | absolute_url }}{{ paginator.next_page }}/">{{ site.data.ui-text[site.locale].pagination_next | default: "Next" }}</a></li>
+      <li><a href="{{ paginator.next_page_path }}">{{ site.data.ui-text[site.locale].pagination_next | default: "Next" }}</a></li>
```
mmistakes: I tested this locally and the `baseurl` doesn't get added to the path properly. I think if you do similar changes you did for the "previous" button link it should fix the problem.

Something like this worked for me:

```html
<li><a href="{{ site.paginate_path | replace: ':num', paginator.next_page | replace: '//', '/' | absolute_url }}">{{ site.data.ui-text[site.locale].pagination_next | default: "Next" }}</a></li>
```

mmistakes:
Thanks for this. 

If you can make the one small change to fix the "Next" link with sites that use `baseurl` and then submit this against the `develop` branch it should be good to merge.

maik:
I've applied your fix and created a new pull request against the develop branch. I guess, we can close this PR now?

mmistakes:
Thanks @maik. I'll close this and take a look at the new PR.

PR #761, opened 2017-01-09
Build sass files for gem to speed up jekyll build time, only include …

mje-nz:
For gem installs, build the stylesheets ahead of time. This brings the `jekyll build` time of a fresh install from about 3.8s to about 0.2s on my machine.

This change does not affect "repo fork" installs.


mmistakes:
Only issue I have is there are users that wish to customize the Sass partials. It's come up a few times already and I'd hate to lose that ability just to shave off a few seconds of build time. 

To be honest the various `for` loops in the theme eat up way more of the build time than preprocessing SCSS assets.

mje-nz:
You still can customize the partials -- if you're using a "repo fork" install this changeset has no effect, as Jekyll still builds the stylesheets itself. With a gem install there'd be an extra step, but no more so than customizing the javascript files in that scenario.

mmistakes:
I get that, but still doesn't help the users who want to override SCSS files when using the theme as a gem. It's come up twice in the last couple of weeks so it's definitely a way people expect to use the theme.

JavaScript I'm less worried about and feel fine bundling in a compiled version as most almost never touch the scripts.

mje-nz:
Oh right, I see your point. Never mind then :)

mmistakes:
Thanks for the PR and discussion though 😀

PR #764, opened 2017-01-11
Pagination now works for index.html files in subfolders. Links to next page consider baseurl now.

maik:


mmistakes:
Sorry it took me awhile to review this. Everything looks good. Thanks for taking the time to get this working and submitting a PR.

Sarafian:
Can someone provide an example of how this works.
I've created a page `_pages/recent-archive.md` like this

```yaml
---
layout: home
permalink: /recent/
author_profile: true
title: Recent Posts
---
```

But the paginator object is `nil`. I've tried to configure the `_config.yaml` with `paginate_path: /recent/page:num/` but it created all other pages except the first one.

mmistakes:
Pretty sure the default Jekyll pagination only works with `.html` files. Rename to `recent-archive.html` and we if that fixes it. 

Sounds like you have everything else configured correctly since you're getting `/recent/page/x/index.html` files generated.

mmistakes:
@Sarafian Scratch that. Forgot about about some other Jekyll pagination limitations.

- Filename has to be `index.html` or else it will be skipped.

So to do what you want rename `_pages/recent-archive.md` to `/recent/index.html` and then rebuild. For whatever reason you can't place pagination indexes inside of a folder like `_pages`.

You'll also want to remove `permalink: /recent/` from that file's YAML Front Matter. Just need the following and everything should work:

```
---
layout: home
author_profile: true
title: Recent Posts
---
```

Sarafian:
It worked indeed as you said. I'll summarize
- Only `index.html` files can reference the `pagination` object.
- When the root is not the desired location then create a folder e.g. `recent` and place the html file inside `recent\index.html`
  - In the `_config.yaml` make sure to set `paginate_path: /recent/page:num/`

Thanks @mmistakes. Once I finish with my setup I'll make check the files/documentation and try to enhance them with what I learned and submit a pull request.

Sarafian:
Strange inconsistency between local and github pages execution
- local works well. The `pagination` object works well out of the `/recent/index.html`. RSS also works well.
- Github repository /sarafian/sarafian.github.io doesn't although all posts are rendered properly.

mmistakes:
Strange. Just looked at your site and seems fine to me. Paginated /recent/ pages are there and populated with posts.

PR #771, opened 2017-01-13
Add https:// prefix to google-universal.html embed code

matthewjberger:
The google universal analytics embed code looks to be slightly different now, including an https before the url to the analytics script.


mmistakes:
Can you submit this PR against the `develop` branch? Thanks.

PR #773, opened 2017-01-15
Translate

JohannEspejo:
Translate to spanish offered.


mmistakes:
Was this a PR to the wrong repo? Not sure I follow what this pull request is for if not. The theme already offers the UI translated into Spanish. 

https://github.com/mmistakes/minimal-mistakes/blob/master/_data/ui-text.yml#L52-L95

PR #783, opened 2017-01-21
Edit docs/configuration (fix analytics example)

vazrupe:
Fix invalid 'analytics' example

[_includes/analytics-providers/google-universal.html#7](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/analytics-providers/google-universal.html#L7)


mmistakes:
Thanks @vazrupe

PR #805, opened 2017-01-30
[plugin] jekyll-data

ashmaroli:
Hi Michael,
The `jekyll-data` plugin is heading towards a `v1.0` release and I would like to refresh the dedicated feature branch.

If you'd delete existing `feature/jekyll-data` branch and re-checkout the same off current `master`, I'll update this PR onto the new branch.

In a nutshell, `jekyll-data` reads data files and a `_config.yml` within theme gems.

This PR does the following:
- [x] modifies existing `test` directory to use and test this plugin.
- [x] modifies existing `Rakefile` to run jekyll commands via the shell instead of the ruby interpreter. (for theme development in future.)
- [x] initializes a travis.yml and a shell script to showcase the plugin in action (and tests in future)
**Requires travis to be enabled on this repository.**
- [ ] update documentation so it would be easier for you to understand.


mmistakes:
@ashmaroli Does `v1.0` get around the need for a user to have to add `jekyll-data` to `gems` in their _config.yml file? While it's not as bad as asking them to copy over any required `_data` files, it is an extra step I have to document.

I was hoping I could just add it as a dependency in the theme's gemspec and it would come along for the ride without any intervention by the user of the theme.

ashmaroli:
No, As of Jekyll v3.4, I'm afraid a plugin can't come along for the ride by just being listed as a dependency.
Its has to be **either** included under `gems:` (in _config.yml) **or** under `:jekyll_plugins` (in Gemfile)

(Yes, this is in direct conflict with what I stated previously in that thread in Jekyll Talk.)

mmistakes:
That's a bummer but since I'm already pushing using a Gemfile with the theme it's probably not the end of the world if they just need to add it under `:jekyll_plugins`. Seems like you documented that well in the PR so I'm cool with that.

I just deleted the existing `feature/jekyll-data` branch and created a new one off of the last `master`. `develop` tends to be more current if you'd rather submit the PR against that, but right now I believe they're the same.

ashmaroli:
I'll keep the PR against the new feature branch as conflicts are bound to arise.

ashmaroli:
Hi,
I'm working on a new idea that'll help **new** Jekyll users to ease into using a loaded theme such as Minimal-Mistakes.
The idea will be released as v0.2.0 of [`jekyll-plus`](https://github.com/ashmaroli/jekyll-plus) in the near future and it will integrate `jekyll-data` with `jekyll` library run with an executable called `jekyll+` and has been built considering how much work needs to be put in by a new user to migrate to `Minimal-Mistakes` starting from `jekyll new`.

Though not-yet-ready-for-release, its base functionality can be taken for a test-drive by following:
  Step 1: install a dummy-gem built off the repo's `master` branch.
  Step 2: install a dummy-version of `minimal-mistakes-jekyll` that includes the `_data` folder and the `_config.yml`
  Step 3: run **`$ jekyll+ new-site test-blog --theme minimal-mistakes-jekyll`**
  Step 4: `$ bundle exec jekyll serve` or `$ bundle exec jekyll+ serve`
  Step 5: Preview site in browser.

-- just 1% short of a fully-functional minimal-mistakes themed site --

Known Issue:
  - comparatively very slow because of requiring multiple gems.

mmistakes:
Nice. I'll give it a spin when I get some free time.

PR #809, opened 2017-02-01
Master

jeffonelson:
Update gem file for theme


mmistakes:
FYI you're pushing to the wrong repo 😉. Update the remote `origin` to point to your repo and that should fix it.

PR #820, opened 2017-02-06
Add medium connector

veelenga:
Sometimes people blog over the internet and want to link it on personal blog pages.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -101,6 +101,7 @@ author:
   steam            :
   tumblr           :
   twitter          :
+  medium           :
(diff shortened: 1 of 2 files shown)


mmistakes:
Thanks for the PR. 

I'm currently not accepting commits of this sort for the author sidebar as it's easy enough to customize and add/remove to meet the users needs. I'm trying to keep this in check to avoid having every imaginable social network option in there.

https://mmistakes.github.io/minimal-mistakes/docs/layouts/#author-profile

PR #834, opened 2017-02-13
Added Korean data in ui-text.yml

varian-lee:
Added Korean translations.

Would be happy if many Koreans could use Minimal Mistakes in Korean!


mmistakes:
Thanks for the PR @varian-lee!

PR #847, opened 2017-02-17
Added a few clarifying comments

ShaneCurcuru:
YAML references are not always obvious, and ensuring people know to remove your Google IDs seemed a good idea to comment on.  Thanks for minimal mistakes!

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -9,7 +9,7 @@
 locale                   : "en-US"
 title                    : "Minimal Mistakes"
 title_separator          : "-"
-name                     : &name "Michael Rose"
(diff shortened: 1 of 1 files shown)


mmistakes:
Thanks.

The "canonical" `_config.yml` [is here](https://github.com/mmistakes/minimal-mistakes/blob/master/_config.yml)... so I wasn't too worried about adding the comments to the documentation's site config.

But still a good addition. I know a few people like to use `/docs` as a starter... going so far as forgetting to remove my Google Adsense banners from their sites :stuck_out_tongue_closed_eyes:

ShaneCurcuru:
Ha! I knew it. And d'oh for patching just the docs/ version.

PR #858, opened 2017-02-19
meta tag "author" added, for site author the field name is used

rsaccani:
I added the meta tag author which was missing.

In case of fallback to site author, seo_author is now assigned to site.author.name instead of site.author[0] which always produced an empty string.


mmistakes:
Looks good. Thanks!

PR #867, opened 2017-02-24
Remove extraneous comments

ShaneCurcuru:
Submitted upstream: https://github.com/mushishi78/jekyll-group-by-array/pull/1

Also check the live output of the 10-layouts page, which seems to be interpreting the code in the "Tags Archive page" section
https://mmistakes.github.io/minimal-mistakes/docs/layouts/

### DIFF
--- a/_includes/group-by-array
+++ b/_includes/group-by-array
@@ -4,44 +4,24 @@
 # © 2015 Max White <mushishi78@gmail.com>
 # MIT License
 -->
-
-<!-- Initialize -->
 {% assign __empty_array = '' | split: ',' %}
 {% assign group_names = __empty_array %}
 {% assign group_items = __empty_array %}
-
-<!-- Map -->
 {% assign __names =  include.collection | map: include.field %}
-
-<!-- Flatten -->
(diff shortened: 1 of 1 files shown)


mmistakes:
Good catch on the Layout page, forgot to escape the Liquid there. Not sure what value removing the comments has though as they don't appear to be extraneous to me.

ShaneCurcuru:
The issue with the comments is they get put in the output HTML - meaning a bunch of extraneous HTML comments bulking up everyone's live static sites.  Just an optimization; I agree comments are good on develop, but in releases they should go IMO.  But up to you. Thanks.

mmistakes:
You can easily remove them with a variety of methods. The theme at one point used the the `compress` layout which strips out comments and white-space. I took it out because there were some edge-cases where it could break JavaScript.

But if its something you want you can easily copy the `compress` layout to your site, add `layout: compress` to `_layouts/default.html` and away you go.

Full instructions here: http://jch.penibelst.de/

ShaneCurcuru:
Got it - thanks for detailed docs on how to do it.

PR #869, opened 2017-02-24
Add AnchorJS. Resolves #860

### LINKED ISSUE #860: AnchorJs - Anchor content headers
Just an idea. Maybe use AnchorJs to create anchor links to content titles?

jhabdas:
Ported from my site and simplified using `defer` attribute. I'll leave some inline notes so you understand why I shuffled some stuff around. Please feel free to use and abuse this branch.

_Sidebar_: It'd be a good idea to use the [Google analytics async script](https://developers.google.com/analytics/devguides/collection/analyticsjs/) too, and move it into the head. It helps performance (especially over HTTP/2).


assets/js/_main.js
```diff
@@ -70,3 +70,17 @@ $(document).ready(function(){
...
+  icon: '\uf02e' // http://fontawesome.io/icon/bookmark/
+}
+anchors.add(
+  '.page__content > h1,' +
+  '.page__content > h2,' +
+  '.page__content > h3,' +
+  '.page__content > h4,' +
+  '.page__content > h5,' +
+  '.page__content > h6'
+)
```
jhabdas: AnchorJS does not use jQuery. But I believe it does require the document to need to be parsed (hence the use of the [`defer` attribute](http://devdocs.io/html/attributes#defer-attribute) on the async script in the head.

_includes/head.html
```diff
@@ -31,4 +31,5 @@
   </style>
 <![endif]-->
 
-<meta http-equiv="cleartype" content="on">
\ No newline at end of file
+<meta http-equiv="cleartype" content="on">
+<script async defer src="{{ '/assets/js/main.min.js' | absolute_url }}"></script>
```
jhabdas: Moves script into head and loads it asynchronously. This allows browsers to start the download as soon as they can (much sooner than before the close of the body) but do so in a non-blocking manner. Big benefits here for hosts supporting HTTP/2, and in some cases modern browsers may perform optimistic preloading as described in the Google analytics doc I linked.

jhabdas: Also note, if you're concerned about anchor supporting older browsers, you can move the anchor JS snippets I dropped into `_main` into a function called `run` and add this to the same file:

```
// Run the application when both DOM is ready and page content is loaded
if (['complete', 'loaded', 'interactive'].includes(document.readyState) && document.body) {
  run();
} else {
  document.addEventListener('DOMContentLoaded', run, false);
}
```

mmistakes:
How does this affect servers not using HTTP/2... i'm guessing most don't. I don't really want to advocate it if it's going to hurt performance for the majority of the users. Especially according to this line:

> Only use this tracking snippet if your visitors primarily use modern browsers to access your site.

Been trying not to open the "over-optimized" can of worms with the theme. I've always viewed it as starting point with just enough best practice nods to push newbies in the right direction. 

If you care about this sort of stuff you're going to hack it up to your liking anyways and likely should be rolling your own theme. Afraid if I start down this path the next thing will be optimizing for the critical path... which would be really hard as a theme since I have no control over the sort of content people are going to load up.

Then there are fonts... and how best to load those... and responsive images via srcset... and a million other things that would really amp up the complexity of using this theme.

jhabdas:
> How does this affect servers not using HTTP/2

93% of the worlds' browsers [will have no problem](http://caniuse.com/#feat=script-async) using `async`/`defer`. In fact, not using it would be in violation of the Pareto principle for both the `_main.js` scripts as well as Google Analytics.

jhabdas:
FYI - Updated last comment to be more explicit and informative. On more thing too, you'll probably need to do some CSS treatments for the links after a little tweaking. This is just an example.

_includes/head.html
mmistakes: Does adding both `async` and `defer` attributes actually work? I was under the impression you had 3 options with scripts.

> There are three possible modes that can be selected using these attributes. If the async attribute is present, then the script will be executed asynchronously, as soon as it is available. If the async attribute is not present but the defer attribute is present, then the script is executed when the page has finished parsing. If neither attribute is present, then the script is fetched and executed immediately, before the user agent continues parsing the page.

Source: https://www.w3.org/TR/html5/scripting-1.html#attr-script-async

jhabdas: It provides [better fallback support](http://stackoverflow.com/a/13821233) for IE. I picked up the pattern from the CloudCannon team. IMO we can ((and always should) expect browsers to make the appropriate choice when both are provided.

jhabdas: Re my IMO, that's because HTML5 put the onus of handling spec exceptions on the browsers. If it weren't that way, we'd have conditional comments and object detection for every browser, and a spec full of exceptions. How fun would that be? :)~

mmistakes: Guess I should have kept scrolling down the page because the W3 said exactly what you did 😛 

> The defer attribute may be specified even if the async attribute is specified, to cause legacy Web browsers that only support defer (and not async) to fall back to the defer behavior instead of the synchronous blocking behavior that is the default.

mmistakes:
Just tried to test this locally and getting this error in the console

```
Uncaught ReferenceError: anchors is not defined at main.min.js:5
```

As far as I can tell `plugins/anchor.js` was added to `main.min.js` and properly initialized.

_includes/head.html
jhabdas: Groovy. Nice to have a more authoritative source to reference in the future. Thanks for sharing.

jhabdas:
@mmistakes Fixing. Uno momento.

jhabdas:
@mmistakes pushed up another commit which should fix the issue. I removed the `defer` attribute as I've introduced a secret sauce script which uniformly detects the page has been parsed across browsers (the reason you were seeing the `undefined` error.

This may look like magic, but it's part of the reason jQuery was invented in the first place. If you want to, you could simply move the contents of the `run` function into the jQuery DOMReady wrapper function and eliminate my JS-fu. I'd've done so myself but cannot, as a good dev, blindly suggest an antipattern without first showing the proper way to do it.

I considered moving the magic into a `utils` folder but `import` (from the ES2015 spec) has [zero percent adoption](http://caniuse.com/#feat=es6-module) in 2017 and would never recommend anyone use Babel or Webpack unless they're building a single-page app.

jhabdas:
Whoops. You'll need to run that through the minifier before running.

jhabdas:
Updated to fix a bug with `use strict` (I forgot to put it in quotes) and uglified the script for ya.

mmistakes:
You might want to branch off of `develop`. `master` doesn't have the latest stuff and it's likely going to be a pain for me to merge this in due to conflicts and such.

jhabdas:
I'll just close for now. Just wanted to provide a sample implementation for you to use if you wanted to add the feature (additional treatments necessary from design perspective).

PR #901, opened 2017-03-15
Remove buggy conditional around related posts header

### LINKED ISSUE #900: Odd condition on "You May Also Enjoy" translated string displaying
- [X] I believe this to be a bug with the theme --- not Jekyll, GitHub Pages or one of the bundled plugins.
- [X] I have updated all gems with `bundle update`.
- [X] I have tested locally with `bundle exec jekyll build`.

---

## Environment informations

- **Minimal Mistakes version:** 4.3.0
- **`github-pages` or `jekyll` gem version:** 3.4.2
- **Operating system:** OS X 10.11.6

---

## Expected behavior

The string "You May Also Enjoy" string or it's translation should always be displayed before related posts section. It should be displayed whether you have a translation file available or not and even if you have a translation file without the matching key (`related_label`).

As of today, if you don't have a translation file (that's how I understood that the translations wasn't bundled with the gem) this string won't display so it will leave the "related posts" section without any description what could be confusing for the reader.

It's because of the conditional available here: https://github.com/mmistakes/minimal-mistakes/blob/master/_layouts/single.html#L66

I think that we don't want this test at all since the line in the condition has a default value.
 
Pretty simple:

- create a new Jekyll site
- add the minimal-mistakes gem
- enable related posts
- add multiple posts
- run `jekyll s`
- go to the latest  post
- check "related posts" section

The related label will be missing.

capture d ecran 2017-03-15 a 10 51 55

Add the translation file with the key and then it displays as intended:

capture d ecran 2017-03-15 a 10 53 36

If it helps I can PR a patch. Oh and BTW you theme is *really awesome*

Bounga:
This removes a buggy conditional that checks if `related_label` is
available in translations and prevents header to be displayed in
some edge cases.

Fix #900

### DIFF
--- a/_layouts/single.html
+++ b/_layouts/single.html
@@ -63,9 +63,7 @@ layout: default
   {% comment %}<!-- only show related on a post page when not disabled -->{% endcomment %}
   {% if page.id and page.related and site.related_posts.size > 0 %}
     <div class="page__related">
-      {% if site.data.ui-text[site.locale].related_label %}
-        <h4 class="page__related-title">{{ site.data.ui-text[site.locale].related_label | default: "You May Also Enjoy" }}</h4>
-      {% endif %}
+      <h4 class="page__related-title">{{ site.data.ui-text[site.locale].related_label | default: "You May Also Enjoy" }}</h4>
       <div class="grid__wrapper">
         {% for post in site.related_posts limit:4 %}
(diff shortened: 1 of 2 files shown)


mmistakes:
Thanks for taking care of this @Bounga.

PR #930, opened 2017-04-04
Replace modified with last_modified_at

### LINKED ISSUE #929: Use last_modified_at instead of or in addition to modified?
- [x] This is a question about using the theme.
- [x] This is a feature request.
- [x] I believe this to be a bug with the theme --- not Jekyll, GitHub Pages or one of the bundled plugins.

---

Minimal Mistakes uses the `modified` property of a page for the ["Updated:" footer in single layout](https://github.com/mmistakes/minimal-mistakes/blob/805dad00775ad5e36b1e8b7ac8d8570242d72ef4/_layouts/single.html#L46).  However, jekyll-feed uses [last_modified_at](https://github.com/jekyll/jekyll-feed/blob/c048229b94c803b1b5d1f5597e2a1820a23977a9/lib/jekyll-feed/feed.xml#L40) which can be set automatically by a [Jekyll plugin](https://github.com/gjtorikian/jekyll-last-modified-at).

So, currently I have to set both "modified" and "last_updated_at" if I'd like to have correct dates on the page as well as in the atom feed.  

Would it be possible for Minimal Mistakes to use "last_updated_at" instead of or in addition to "modified"?

lunaryorn:
Fixes #929

### DIFF
--- a/_layouts/single.html
+++ b/_layouts/single.html
@@ -21,7 +21,7 @@ layout: default
     {% if page.title %}<meta itemprop="headline" content="{{ page.title | markdownify | strip_html | strip_newlines | escape_once }}">{% endif %}
     {% if page.excerpt %}<meta itemprop="description" content="{{ page.excerpt | markdownify | strip_html | strip_newlines | escape_once }}">{% endif %}
     {% if page.date %}<meta itemprop="datePublished" content="{{ page.date | date: "%B %d, %Y" }}">{% endif %}
-    {% if page.modified %}<meta itemprop="dateModified" content="{{ page.modified | date: "%B %d, %Y" }}">{% endif %}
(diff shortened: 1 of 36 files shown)


mmistakes:
Fantastic! Thanks for this.

lunaryorn:
@mmistakes Thank you for merging, and for this awesome theme.

Would you mind to publish a new release of this theme to rubygems including this change?

mmistakes:
@lunaryorn I'm not ready to roll out a new release but you can replace `gem "minimal-mistakes-jekyll"` with the following in your `Gemfile`.

```ruby
gem "minimal-mistakes-jekyll", :github => "mmistakes/minimal-mistakes"
```

 It will build the gem for you using the latest on `master`. It's a nice trick if you want to be current without waiting for an actual gem release on Rubygems.

lunaryorn:
Ah, thanks! ❤ Didn't know that 👍

PR #941, opened 2017-04-09
sidebar: added support for html content

rsaccani:
Hi, I've added the possibility to enter html content in the custom sidebar.

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -14,6 +14,7 @@
       {% endif %}
       {% if s.title %}<h3>{{ s.title }}</h3>{% endif %}
       {% if s.text %}{{ s.text | markdownify }}{% endif %}
+      {% if s.html %}{{ s.html }}{% endif %}
     {% endfor %}
     {% if page.sidebar.nav %}
       {% include nav_list nav=page.sidebar.nav %}


mmistakes:
Not sure if this is needed. You can already add HTML using `text`. It accepts Markdown which also should accept HTML since Kramdown processes both.

Does this not work?

```html
sidebar:
  - title: "Title"
    image: http://placehold.it/350x250
    image_alt: "image"
    text: "<strong>Here's some HTM.</strong>"
```

rsaccani:
I see. I had problems adding a form (paypal button) using sidebar.text, that's why I made this change. 
Probably it was just a multiline formatting issue. I just tried again with sidebar.text putting everything in a single line and it worked fine.
So, it looks like there's no need for sidebar.html.

mmistakes:
If you use literal line scalar in YAML (e.g. `|`)  multiline HTML should work just fine. I tested this with no issues.

```yaml
sidebar:
  - title: "Another Title"
    text: |
      <table>
        <tr>
          <td>Sample cell 1</td>
          <td>Sample cell 2</td>
        </tr>
      </table>
```

RichStone:
Is there a way to include an HTML piece from the _includes folder using the sidebar?

mmistakes:
@RichStone If you need that sort of flexibility you'll need to hack the theme. I'm pretty sure there is no way to pull in a Jekyll include via YAML Front Matter... at least not without some added logic to the theme.

PR #945, opened 2017-04-10
fix(includes/video): use https always

### LINKED ISSUE #944: Vimeo fails to play when served over HTTPS
https://github.com/mmistakes/minimal-mistakes/pull/788 introduced video URLs with hard-coded scheme. These need to be made protocol-relative for proper support over HTTPs.

jhabdas:
when served over https the vimeo link was failing. defer scheme used for video iframe src urls
realtive to the current protocol (e.g. http or https)

closes #944

### DIFF
--- a/_includes/video
+++ b/_includes/video
@@ -4,7 +4,7 @@
 <!-- Courtesy of embedresponsively.com //-->
 <div class="responsive-video-container">
 {% if video_provider == "vimeo" %}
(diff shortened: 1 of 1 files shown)


jhabdas:
Updated to use HTTPs by default. My understanding is that HTTPS sites will not pass referrer info back to non-HTTPs. If so, oh well? 😺

Issue #883, opened 2017-03-02
Justified text by default


juliohm:
First of all, thank you very much for this awesome project, it is incredible :+1: 

I am trying to make the contents of the posts and pages all justified by default, but I am very new to Jekyll and not an expert in web programming.

How would you change the configuration files for adding a `{: text-justify}` everywhere in the body of the pages?

mmistakes:
You'll have to modify the theme's CSS to do this. If you set the following declaration it will justify the post/page content.

```css
.page__content {
  text-align: justify;
}
```

juliohm:
Thank you @mmistakes , what is your suggestion about the file where I should add these lines? I am seeing many files inside of `_sass`, is that the right place?

mmistakes:
Yes, that's the right place. Here would work https://github.com/mmistakes/minimal-mistakes/blob/master/_sass/_page.scss#L53

juliohm:
Awesome @mmistakes, problem solved. Closing the issue.

YannisPap:
Hello @mmistakes and thanks for this awesome theme.
I tried the above but the right sidebar is part of the page__content and sometimes creates this ugly result:
![fraud detection using machine learning analysis just another data science blog](https://cloud.githubusercontent.com/assets/10160052/24951961/bab34c1e-1f7e-11e7-87c8-c546ab0c1916.png)

Is there a way to exclude the sidebar?

Thank you.

mmistakes:
@YannisPap Sure... with more CSS.

```css
.sidebar__right {
  text-align: left;
}
```

YannisPap:
You are right. Very nubby on CSS.

Thanks you.

badouralix:
Would it be relevant to make a PR adding a new variable in [_sass/_variables.scss](https://github.com/mmistakes/minimal-mistakes/blob/eaf366e405a0f988fa90cd7fe0bc136df60ea34b/_sass/_variables.scss) setting the default alignment, instead of manually patching the theme when one wants to override it?

NickleDave:
In case anyone else comes upon this issue ...

I was able to achieve what's described here
https://github.com/mmistakes/minimal-mistakes/issues/883#issuecomment-283648056
by adding it in `/* paragraph indents */` in the current `_page.scss`

https://github.com/mmistakes/minimal-mistakes/blob/8a67ce8e41ec850f2d7c373aa47739b2abfee6f1/_sass/minimal-mistakes/_page.scss#L121
```
[...]

PR #954, opened 2017-04-16
Add sass variable to set default page content alignment

badouralix:
This pull request comes from #883 and gives a way to choose the default text alignment without having to patch several files of the theme.

### DIFF
--- a/_sass/_page.scss
+++ b/_sass/_page.scss
@@ -59,6 +59,7 @@
 
   p, li, dl {
     font-size: 1em;
+    text-align: $paragraph-align;
   }
 
   /* paragraph indents */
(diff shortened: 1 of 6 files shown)


badouralix:
Broken by [commit 3fc1bfd](https://github.com/mmistakes/minimal-mistakes/commit/3fc1bfde26e09d6b150a5a359144c427d80bbd5a). I'm closing this PR for now, but I still think that the default alignment is worth being set by a variable.

mmistakes:
Trying to avoid these sorts of customization because they lead to more customizations of this nature. It's a simple 1 line CSS declaration so I don't feel too bad not including it as a SCSS variable.

That and in good conscious I can't provide the option for justified type as it almost always looks awful on the web :wink:

badouralix:
Oh okay I see, this is a good point 👍

Issue #554, opened 2016-09-25
Related posts not shown


fa-ribeiro:
- [ ] This is a question about using the theme.
- [x] I believe this to be a bug with the theme --- not Jekyll, GitHub Pages or one of the bundled plugins.
- [ ] This is a feature request.
- [x] I have updated all gems with `bundle update`.
- [x] I have tested locally with `bundle exec jekyll build`.

---
## Environment informations
- **Minimal Mistakes version: 4.0.1 (feature/theme-gem branch)** 
- **`jekyll` gem version: 3.2.1 (master branch)** 
- **Operating system: Mac OSX 10.11.6** 

---
## Expected behavior

Related post should appear on every post as they are enabled by default in my `_config.yml`:

```
# Defaults
defaults:
  # _posts
  * scope:
      path: ""
      type: posts
    values:
      related: true
```
## Steps to reproduce the behavior
1. create a couple of posts
2. enable related posts on `_config.yml`
## Additional information

If I insert the following code block (extracted from `_layouts/single.html`) in a post, the related posts are shown as expected:

```
{% if page.id and page.related and site.related_posts.size > 0 %}
  <div class="page___related">
    {% if site.data.ui-text[site.locale].related_label %}
      <h4 class="page__related-title">{{ site.data.ui-text[site.locale].related_label | default: "You May Also Enjoy" }}</h4>
    {% endif %}
    <div class="grid__wrapper">
      {% for post in site.related_posts limit:3 %}
        {% include archive-single.html type="grid" %}
      {% endfor %}
    </div>
  </div>
{% endif %}
```

Any ideas?
Thanks in advance for any help!

mmistakes:
Link to a repo so I can troubleshoot? Most likely it's a content or config problem on your end.

fa-ribeiro:
it's a private repo, but here it is the `_config.yml`:

# Welcome to Jekyll!
    #
    # This config file is meant for settings that affect your entire site, values
    # which you are expected to set up once and rarely need to edit after that.
    # For technical reasons, this file is *NOT* reloaded automatically when you use
    # `jekyll serve`. If you change this file, please restart the server process.

    theme                    : "minimal-mistakes-jekyll"

    # Site Settings
    locale                   : "pt-PT"
    title                    : "FRibeiro.org"
    title_separator          : "-"
    name                     : "Fernando Ribeiro"
    description              : "Random thoughts, quick notes, some stories and a few ideas."
[...]

PR #978, opened 2017-04-27
Replace `site.related_posts` with Liquid loops for improved relevancy

mmistakes:
`site.related_posts` currently outputs an array of the most recent posts for GitHub Pages hosted sites since `LSI` is disabled, as described in #554.

This PR improves on it with some Liquid code that tries to improve the relevancy of these "related posts" by comparing them with `page.tags`.

Note: there may be a slight increase in build times as all of the posts need to be looped through, which can obviously slow things down a tad if you have many posts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Enhancements
+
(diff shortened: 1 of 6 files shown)


mmistakes:
Closed as this negatively impacts build times.

PR #980, opened 2017-04-29
Master

sambrentnall:
- Updated config files
- Created home.md with header
- Created sample posts

Next
Homepage layout


mmistakes:
FYI you're pushing to the wrong remote repo. Check that your Git `origin` is set properly.

sambrentnall:
Thanks, still learning Git and think I've sorted it now.

PR #1038, opened 2017-05-29
Removing extraneous a and li tags

mtlynch:
Removing duplicate <a> and <li> closing tags in paginator include.


mmistakes:
Thanks @mtlynch

PR #1054, opened 2017-06-06
Fixes and updates for the Greek language

hkalant:
I have translated some UI text that hadn’t been translated by @Stelios3g. And I have also changed some translated entries to look and sound more natural for general uses (eg ‘Recent posts’ was translated as ‘Latest Announcements’ before). Finally, I have added ‘RSS’ next to ‘Feed’, because 1) there is no translation (or something that sounds natural) in Greek I’m afraid, 2) the term ‘RSS Feed’ is widely used in Greek websites.


mmistakes:
@Stelios3g care to weigh in?

As far as adding RSS, the feed is technically Atom and not RSS. People commonly interchange them but they're two different specs which is why I went with "Feed".

hkalant:
Hi, Michael.
I am Greek myself and I can assure you that the translation is correct.
Please feel free to keep 'Feed' in English, because there is no term in Greek that could describe such thing and isn't related to 'food/eating'.
Many thanks.

PR #1069, opened 2017-06-18
Fix broken link to example _config.yml file

annegentle:
Configuration docs were pointing to example `_config.yml` on a gh-pages branch which showed Not Found when clicked.


mmistakes:
Good catch. Thanks.

PR #1079, opened 2017-06-24
Mistake on file navigation.yml

leggiero:
Previous wrong name was `navigations.yml`


mmistakes:
Thanks for the typo fix.

PR #1087, opened 2017-06-28
Update head.html

justinrummel:
Conditional comment as this is a fix for IE Mobile devices.  By doing this resolves validation errors.  

Same as: https://github.com/h5bp/mobile-boilerplate/commit/feb59c52ae77bb80a2a76d2796cf56e0c4a71c98

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -31,4 +31,4 @@
   </style>
 <![endif]-->
 
-<meta http-equiv="cleartype" content="on">
+<!--[if IEMobile]> <meta http-equiv="cleartype" content="on"> <![endif]-->


mmistakes:
Do you know if this affects desktop browsers that could benefit from turning on cleartype? And if the line even needs to be there anymore? Looking at that commit from the mobile H5BP, that project was deprecated years ago.

And in the main H5BP they don't use the cleartype meta line at all.

justinrummel:
Unfortunately I'm not a knowledgeable resource on this item.  I was running my site through some validators (Rocket Validator specifically) and this item surfaced.  I thought the general site could use this update vs. creating a local _includes/head.html gem override.  I did question if the right course of action would to remove the line entirely... but again I'm not sure.

I don't know if this effects all other browsers... I've only tested on Desktop and Mobile Safari,  everything looks fine.

mmistakes:
I'm not too worried about it failing some validation as the W3C validator
already flags a few things. In my experience there's always a few things
that will trip it up which have no impact on the user.

Thinking I'll leave it in as is... with the expectation to remove it
eventually. I couldn't find much documentation on it other than people
complaining about blurry fonts a few years ago. Which who knows if its an
issue now a days with modern browsers.

PR #1088, opened 2017-06-29
 Add $navicon-link-color-hover 2/2

palatinux:
Since the orig color-scheme doesn't always represent the nav/site colors


mmistakes:
@palatinux Could you combine these two commits into a single PR?

PR #1089, opened 2017-06-29
Add $navicon-link-color-hover 1/2

palatinux:
Since the orig color-scheme doesn't always represent the nav/site colors


palatinux:
You can compare now with https://github.com/palatinux/minimal-mistakes/ Don't know how to push multiple commits into one push

mmistakes:
The easiest way is to fork, create a new branch from `master`, add your commits, push that branch up, then submit a PR.

You can also keep pushing commits to this branch, though looking up above looks like you might have deleted the source branch.

It's a cleaner merge if it's all in one PR, something this small doesn't need to be broken up into multiple PRs.

PR #1131, opened 2017-07-19
Use a plugin to read data files from gem

ashmaroli:
Hello Michael,
Revisiting this again. Jekyll learnt a few new tricks since the last time I came with a proposal. It will now automatically require a theme's runtime dependencies. Ergo, the end-user doesn't have to add the required plugins to their config file.

Simply having `theme: minimal-mistakes-jekyll` will call the dependencies as if it were a Ruby program

With `jekyll-data`, users need not update their data files with the changes you make. It'll be pulled in from the theme-gem itself.
(It's a different story, if users prefer to use custom ui-text strings though)

With this PR
  - jekyll-data v1.x
  - jekyll 3.5 automatically requires a theme's runtime deps
  - jekyll v3.5.0 had bugs. Hence use v3.5.1 and beyond


mmistakes:
Nice! Thanks for the PR. When I get a chance I just want to give it a quick test and then I think it'll be easily merged in.

ashmaroli:
Thank you Michael :smiley:

PR #1138, opened 2017-07-24
Header alt tags

austinseraphin:
This allows specifying a custom alt tag for a header image. Currently the theme defaults to the site title or nothing. I use a screen reader so noticed this right away, and fixed it on my site, so thought I'd push the change. Simply specify the image_description along with the image in the front matter.


mmistakes:
Thanks for the PR @austinseraphin

Issue #1146, opened 2017-08-02
Overriding global-transition sass variable does not disable page animations


peterdimou:
- [x] This is a question about using the theme.
- [x] I have updated all gems with `bundle update`.
- [x] I have tested locally with `bundle exec jekyll build`.
- [x] I believe this to be a bug with the theme --- not Jekyll, GitHub Pages or one of the bundled plugins.

---

## Environment information

- **Minimal Mistakes version:** 4.4.2
- **`github-pages` or `jekyll` gem version:** `jekyll` 3.5.1
- **Operating system:** macOS 10.12.6

---

## Expected behavior

Disabled page animations for the theme.

## Steps to reproduce the behavior

1. Use gem based version of theme
2. Copy file `assets/css/main.scss` from gem bundle to local site repo (preserving directory structure)
3. Make the contents of the file the following:
```
---
# main.scss
---

@charset "utf-8";

/* Theme overrides */
$global-transition : none;

@import "minimal-mistakes";
```
4. Run `jekyll serve`
5. Navigate to `localhost:4000` and observe page animations

mmistakes:
The `$global-transition` variable isn't used everywhere, so setting it to `none` won't disable all animations.

For example the page fade in animation is applied to the `#main` element and doesn't use the global transition.

```scss
#main {
  @include container;
  @include clearfix;
  margin-top: 2em;
  padding-left: 1em;
  padding-right: 1em;
  animation: intro 0.3s both;
  ...
}
```

You'll need to override all of those as well. Unless someone wants to create a pull request that swaps `animation: intro 0.3s both` with a new Sass variable that can be set...

```scss
$intro-transition: intro 0.3s both;

#main {
  animation: $intro-transition;
  ...
}
```

PR #1147, opened 2017-08-02
Refactor intro animations into a separate sass variable

peterdimou:
Addresses #1146 .

Tested locally, everything seemed to work as expected. Let me know if I need to change anything.


mmistakes:
Works great. Adding the following to `assets/css/main.scss` disables the fade in animations as expected.

```scss
---
# Only the main Sass file needs front matter (the dashes are enough)
---

@charset "utf-8";

$intro-transition: none;
@import "minimal-mistakes";
```

Thanks for taking the time to submit the PR and address the issue @peterdimou

PR #1150, opened 2017-08-03
Include documentation on disabling animations

peterdimou:
Just thought I'd include a section in the docs about it since it's now officially supported and it might be a common enough use case to warrant explanation.

### DIFF
--- a/docs/_docs/16-stylesheets.md
+++ b/docs/_docs/16-stylesheets.md
@@ -45,7 +45,7 @@ To override the default [Sass](http://sass-lang.com/guide) (located in theme's
 
 1. Copy directly from the Minimal Mistakes theme gem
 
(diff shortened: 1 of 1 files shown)


mmistakes:
Good idea. Thanks for taking the time to beef up the docs.

docs/_docs/16-stylesheets.md
```diff
@@ -45,7 +45,7 @@ To override the default [Sass](http://sass-lang.com/guide) (located in theme's
 
 1. Copy directly from the Minimal Mistakes theme gem
 
-   - Go to your local Minimal Mistakes gem installation directory (run 
+   - Go to your local Basically Basic gem installation directory (run
```
ashmaroli: Return of the typo monster

PR #1156, opened 2017-08-05
Add a few more social networks

ducthienbui97:
I just added a few more social networks for author profile: twitch, quora, vk, reddit, yelp, meetup, angellist, paypal (donate link), skype (open chat),  and added color for weibo logo.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -83,6 +83,7 @@ author:
   location         : "Somewhere"
   email            :
   uri              :
+  angellist        :
   bitbucket        :
   codepen          :
   dribbble         :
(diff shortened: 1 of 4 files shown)


halirutan:
Please see this comment here: https://github.com/mmistakes/minimal-mistakes/pull/1142#issuecomment-318939268

mmistakes:
Thanks for the PR @ducthienbui97, but as @halirutan linked to... I'm not accepting these sorts of adds. It is easy enough to customize I don't think it's productive to add every possible social network to the author sidebar.

It'll be never ending, and never quite match the order someone expects them to be listed in.

PR #1177, opened 2017-08-15
URL Encode Title and Page URL in Social Share Links

rjkaes:
The links to the various social networks include data from the post
(title and url) which can contain URL special characters.  Pass each of
the generated URLs through Liquid's `url_encode` filter to build HTML
complaint URLs.


mmistakes:
Thanks @rjkaes

PR #1181, opened 2017-08-16
Fix discourse urls

tendon:
The current discourse embedding snippet was not working for me. I found two problems

1) the `discourseUrl` variable hardcoded  a couple of forward slashes at the beginning. The correct path should be set only with the `site.comments.discourse.server` configuration variable.

2) the canonical url assigned to `discourseEmbedUrl` was not computed correctly.

This commit attempts to fix both.

### DIFF
--- a/_includes/comments-providers/discourse.html
+++ b/_includes/comments-providers/discourse.html
@@ -1,7 +1,7 @@
 {% if site.comments.discourse.server %}
-{% capture canonical %}{% if site.permalink contains '.html' %}{{ page.url | absolute_url }}{% else %}{{ page.url | absolute_url | remove:'index.html' | strip_slash }}{% endif %}{% endcapture %}
+{% capture canonical %}{{ page.url | replace: 'index.html','' | prepend: site.baseurl | prepend: site.url }}{% endcapture %}
 <script type="text/javascript">
-    DiscourseEmbed = { discourseUrl: '//{{ site.comments.discourse.server }}/',
+    DiscourseEmbed = { discourseUrl: '{{ site.comments.discourse.server }}/',
                        discourseEmbedUrl: '{{ canonical }}' };
    (function () {
      var d = document.createElement('script'); d.type = 'text/javascript'; d.async = true;


rriemann:
My two cents:

- I use currently still a very old version of jekyll+this theme, i.e. I have no experience here.
- I could imagine that the discourse canonical url must match the url in the RSS feed that is read in by Discourse. So shouldn't we create the canoncical url in the same manner like in feed.xml ?

tendon:
Yes looking at the canonical url in xml feed makes sense. Where is that
constructed?

mmistakes:
@tendon That's handled not by the theme but by [jekyll/jekyll-feed plugin](https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml).

rriemann:
It seems that feed.xml uses simply

    {{ post.url | absolute_url }}

_includes/comments-providers/discourse.html
```diff
@@ -1,7 +1,7 @@
 {% if site.comments.discourse.server %}
-{% capture canonical %}{% if site.permalink contains '.html' %}{{ page.url | absolute_url }}{% else %}{{ page.url | absolute_url | remove:'index.html' | strip_slash }}{% endif %}{% endcapture %}
+{% capture canonical %}{{ page.url | replace: 'index.html','' | prepend: site.baseurl | prepend: site.url }}{% endcapture %}
```
mmistakes: Instead of prepending `baseurl` and `url` use the `absolute_url` filter which does both of these operations.

mmistakes:
Which is pretty much what we're using now, just with a few filters to deal with `index.html` in paths.

_includes/comments-providers/discourse.html
tendon: Maybe I misconfigured something, but I am pretty sure the `canonical` resulted in a relative url, and that's why I got here. Go figure.
So would this be the simpler solution?:
```
{{ page.url | replace: 'index.html','' | absoulte_url }}
```

mmistakes: Sure, but that's pretty much what it is now... except your version removes a few more checks.

I guess I'm waiting for someone else to verify this is actually broken and not outputting the expected URL for Discourse and not some sort of Jekyll/environment/configuration issue.

PR #1190, opened 2017-08-19
Update Gemfile

dontito94:


ashmaroli:
@dontito94 This change is unnecessary. Jekyll 3.5 and above will automatically load these plugins from the `gemspec`.

(I recommend installing Bundler gem if you don't have it, and then build/serve as following:)

```sh
bundle exec jekyll build
# or serve
bundle exec jekyll serve
```

Additionally, GitHub Pages won't look at this Gemfile. If you're using the repo-fork method with this theme, simply ensure that your config file lists the necessary plugins.

mmistakes:
@dontito94 The branch you're committing this to is old and outdated. I'm just keeping it around for reference purposes for those who may be using version 2 of this theme. Everything @ashmaroli suggested is the preferred way to use the theme.

PR #1195, opened 2017-08-22
Do not print the comments form

lsolesen:


mmistakes:
I'd prefer not to hide comments as they can be considered "content" in a lot of cases.

Also, the preferred method to hiding would be to add the element you'd like to hide to [`_sass/minimal-mistakes/_print.scss`](https://github.com/mmistakes/minimal-mistakes/blob/master/_sass/minimal-mistakes/_print.scss#L220-L237).

lsolesen:
@mmistakes This is only hiding the comments form, not the comments themselves. I can revise to use markup instead, if you prefer.

mmistakes:
Gotcha. Yes, please add the form class to `_print.scss` to hide instead of adding a `no-print` class to the HTML.

Issue #803, opened 2017-01-28
Add support for Staticman new comment notifications + nested comments


alecsatin:
- [x] This is a question about using the theme.
- [x] This is a feature request.

Have been learning so much from your theme.   Thank you for the hard work you have put into it - and for the excellent documentation.  

Static man has added the possibility of a "subscription to comments" option via mailgun.  eduardoboucas describes it here: https://github.com/eduardoboucas/staticman/issues/42#issuecomment-262938831

Basically it requires the staticman yml entries to be removed from the _config.yml file and placed in a dedicated staticman.yml.  

How hard would it be to get this functionality to work with minimal-mistakes?

Thank you.

mmistakes:
Great to hear that :smile:.

Adding support for comment reply notifications along with nested replies is on my radar, but I haven't figured out how I want to handle it yet. To use both of those there are some changes I'd need to manage as to not break support for those who don't upgrade to Staticman `v2`.

Essentially it comes down to moving all the Staticman configs out of `_config.yml` and into a `staticman.yml` file and changing the `POST` url in the comment form.

I've done both of these on my personal site and even [wrote about all the changes](https://mademistakes.com/articles/improving-jekyll-static-comments/). If you're interested that might be a good reference to start with until I can bake it into the theme officially.

You might want to take a look at @justinrummel's [repo](https://github.com/justinrummel/jr.com-mm) for inspiration since he's currently trying to use Staticman `v2` features with Minimal Mistakes.

justinrummel:
I'm really close now. Able to POST and display comments by integrating Staticman and Travis using v2. Now need to work on threaded replies. 

My findings is patience is the key. And browser cache will give you false negatives.

alecsatin:
Thank you @justinrummel and @mmistakes . Appreciate the suggestions.  Will continue to investigate and work on this when the site is linked to a custom domain.

justinrummel:
I've got it! Parent/child is now displaying correctly and using Staticman and Travis. 

Next is Mailgun, but that feature is low priority. 

Public Test URL: https://www.justinrummel.com/macworld-2010-pictures/

mmistakes:
@justinrummel RE: working on pull request to add support for Staticman v2. Off the top of my head here's some thoughts/suggestions:

- [ ] Remain backwards compatible with `v1` Staticman implementation
- [ ] Add a new comment provider to _config.yml, perhaps named `staticman-v2`
[...]

PR #1202, opened 2017-08-28
Fix staticman indentation in _config.yml

ohadschn:
Should be under `comments:`

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -28,22 +28,22 @@ comments:
     appid                :
     num_posts            : # 5 (default)
     colorscheme          : # "light" (default), "dark"
-staticman:
-  allowedFields          : ['name', 'email', 'url', 'message']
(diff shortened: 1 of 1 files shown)


mmistakes:
The current indentation is correct. It isn't supposed to be indented under `comments`. Doing so will break things as the theme uses `site.staticman` not `site.comments.staticman`

ohadschn:
Is `staticman` used for non-comment features? If not, I would argue that the indentation is correct but the usage (`site.staticman`) is not..

mmistakes:
It is. Staticman needs it to be `site.staticman` for it to properly read its config settings. 

Way around that is moving to v2 which takes the Staticman settings out of `_config.yml` and into its own dedicated file, `staticman.yml`.

There are some breaking changes in doing that, see #803

PR #1211, opened 2017-08-31
Improve translation for minute_read.

DDoSolitary:
Although the original one can be understood, it seems strange for a native Chinese speaker. I think the updated version should be better.

### DIFF
--- a/_data/ui-text.yml
+++ b/_data/ui-text.yml
@@ -322,7 +322,7 @@ zh: &DEFAULT_ZH
   toc_label                  : "在本页上"
   ext_link_label             : "直接链接"
   less_than                  : "少于"
-  minute_read                : "分钟 阅读"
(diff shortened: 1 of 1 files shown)


mmistakes:
How bad is the original one? I get a lot of these requests to change translated text for a given locale and it becomes a battle with people constantly submitting PRs to change it back and fourth.

The `ui-text.yml` data file is completely editable by the user. My intention was to provide a base line of texts and then its up to the user to tailor those to suit their site/needs.

DDoSolitary:
I believe that any native Chinese speaker would say that the updated version is better (though not the best, I always fail my Chinese exams XD). But it is really difficult for me to explain how is it better.

I love your project and I just wanted to make this project better. At the same time, I understand that it's somewhat annoying to deal with such translation problems. It is up to you to decide whether to merge this PR.

PR #1221, opened 2017-09-04
Add docker environment

alcalyn:
I'm using this Docker environment to work with this Jekyll theme.

I'm sharing it in this PR.

You should be able to test it by fetching this branch, and by running `docker-compose up`.

Then you should access to your fresh install at http://0.0.0.0:4000

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,11 @@
+version: '2'
+
+services:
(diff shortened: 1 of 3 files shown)


alcalyn:
Yep I think that, as it could be used in any jekyll project, it has not its place here.

PR #1232, opened 2017-09-09
Enable Liquid and Front Matter strict mode

ohadschn:
I am a great believer in strict compilation and "warnings as errors" in basically anything that compiles.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -245,3 +245,8 @@ defaults:
       comments: # true
       share: true
       related: true
+
+# Jekyll compilation
+strict_front_matter: true
+liquid:
+ error_mode: strict


mmistakes:
Too opinionated. I'd rather not force this config onto anyone.

ohadschn:
Fair enough. Can you somehow enable it on your personal builds so that compilation warnings don't creep in in the future?

mmistakes:
I did last week. It's [enabled in the test site](https://github.com/mmistakes/minimal-mistakes/blob/master/test/_config.yml#L146-L148) used when developing the theme with `bundle exec rake preview`.

Issue #1238, opened 2017-09-10
Include jQuery above page__footer to allow for upgrade-safe custom JavaScript


ohadschn:
Suppose I want to add some JS code that runs before _main.cs (in my case, I want to add some classes to elements to be processed by `_main.js`, specifically the `image-popup` class: https://github.com/mmistakes/minimal-mistakes/issues/1235).

So I add a JS file to `assets/js` and reference it from `footer/custom.html`, but then I don't have jQuery because it's bundled inside `main.min.js` which is only included after (lower in the HTML).

I realize I could [override the scripts](https://mmistakes.github.io/minimal-mistakes/docs/javascript/) but that would interfere with Gem upgrades.

ohadschn:
Also, getting jquery (and any other external dependency really) from a CDN rather than minifying it inside `main.min.js` would probably increase performance (and reduce hosting costs), both due to the CDN hosting and browser download parallelization.

mmistakes:
Just override `_includes/scripts.html`

There's barely anything in that file and is very likely not to change making it safe. 

As far as pulling out jQuery into a CDN version. Not going to get into that debate. There's to many Stackoverflow and blog posts arguing the benefits of doing it and not.

ohadschn:
Override `_includes/scripts.html` and do what? I can already have my script included before `main.min.js` using `footer/custom.html` as I mentioned above. The problem is that I don't have jQuery. I suppose I could include it again (before my script) but that seems rather wasteful...

Regarding CDN, I took a quick look in SO in it doesn't seem like much of a debate... the [overwhelming](https://stackoverflow.com/a/3832498/67824) [majority](https://stackoverflow.com/a/3832530/67824) of [answers](https://stackoverflow.com/a/21099939/67824) and upvotes are clearly in favor of a CDN.

It is [especially recommended for HTTP/2](https://stackoverflow.com/a/39197383/67824) due to caching improvements.

mmistakes:
Add your scripts after `main.min.js` which bundles jQuery with a few lines of JavaScript to init various plugins (e.g. lightbox).

I'm screwed either way I do it. Include the CDN and someone will want to self host to avoid an external dependency. Bundle it and someone will want a CDN ref'd one. I've done it both ways in the past and it's a losing battle.

I think you'll see this sort of things with themes in general. They can't possibly do ever suggested web performance suggestion and balance features/customization at the same time.

[...]

PR #1241, opened 2017-09-11
Allow `<head>` and footer scripts to be changed via config

mmistakes:
Common feature request is an easy way to add scripts to `<head>` or in the footer without modifying any of the theme files.

Currently this can be done by adding custom `script `elements to [`_includes/head/custom.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/head/custom.html) or [`_includes/footer/custom.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/footer/custom.html) This pull request allows the user to add additional scripts to either `<head>` or before the closing `</body>` elements via their `_config.yml` file.

Simply add the full relative (or absolute) path(s) to relevant array and they will be added to the page.

Examples:

```yaml
head_scripts:
  - https://code.jquery.com/jquery-3.2.1.min.js
  - /assets/js/your-custom-head-script.js
footer_scripts:
  - /assets/js/your-custom-footer-script.js
```

In the case of `footer_scripts`, if used, the default `/assets/js/main.min.js` file won't be loaded and jQuery and other scripts used by the theme will need to be added to the array or replacements found. Unless you want the collapsible navigation, image lightboxes, smooth scrolling, and other niceties to stay broken ;-)

**Note:** depending on the size of your site there may be a slight performance hit to build times as each of these config variables introduce `for` loops.

Ref: #1238


_includes/head.html
```diff
@@ -29,4 +29,15 @@
       height: 100%;
     }
   </style>
-<![endif]-->
\ No newline at end of file
+<![endif]-->
+
+{% if site.head_scripts %}
+  {% for script in site.head_scripts %}
+    {% if script contains "://" %}
```
ohadschn: Maybe just "//" to allow for protocol-relative URLs (below too): https://stackoverflow.com/questions/4831741/can-i-change-all-my-http-links-to-just/27999789

mmistakes: Protocol less URLs are now considered an anti-pattern. It's preferable to use `https` when available.

https://www.paulirish.com/2010/the-protocol-relative-url/

ohadschn:
Good stuff! So a custom `footer_scripts` with functionality  equivalent to the out-of-the-box one would look like this (going by the [uglify script](https://github.com/mmistakes/minimal-mistakes/blob/dc47f82af226eed9ed42a8daab54101aa14b9c12/package.json))?

```yaml
footer_scripts:
  - /assets/js/vendor/jquery/jquery-1.12.4.min.js
  - /assets/js/plugins/jquery.fitvids.js
  - /assets/js/plugins/jquery.greedy-navigation.js
  - /assets/js/plugins/jquery.magnific-popup.js
  - /assets/js/plugins/jquery.smooth-scroll.min.js
  - /assets/js/_main.js
```

_includes/head.html
ohadschn: I know, that's exactly what I answered on the very same SO thread :) Just thought you'd might want to support it, but I suppose there's really no need, any CDN today would have HTTPS...

mmistakes:
Looks about right. That should be everything.

_includes/head.html
mmistakes: Exactly. And I'm not sure if `//` might filter out something it shouldn't be. I doubt anyone would have a path with that, but seems possible.

ohadschn: Yeah I had the same concern, theoretically `https://github.com/mmistakes//minimal-mistakes` works too. I think even `https://github.com/mmistakes://minimal-mistakes` could work but since Liquid doesn't contain any `startswith` method that I can tell probably not worth the trouble.

ohadschn:
Great, so now anyone could do this to get a CDN-equivalent:

```yaml
footer_scripts:
  - https://cdnjs.cloudflare.com/ajax/libs/jquery/1.12.4/jquery.min.js
  - https://cdnjs.cloudflare.com/ajax/libs/fitvids/1.2.0/jquery.fitvids.min.js
  - /assets/js/plugins/jquery.greedy-navigation.js (didn't see a CDN for this)
  - https://cdnjs.cloudflare.com/ajax/libs/magnific-popup.js/0.9.9/jquery.magnific-popup.min.js
  - https://cdnjs.cloudflare.com/ajax/libs/jquery-smooth-scroll/1.7.2/jquery.smooth-scroll.min.js
  - /assets/js/_main.js
```

BTW, regarding the versions:
* jQuery - I guess you're using 1.x for IE6-8 support? 
* jquery.fitvids.js - latest is 1.2.0 (MM is using 1.1)
* jquery.greedy-navigation.js - looks like a couple of small updates have been made (I diffed the MM version with the one currently on master)
* magnific-popup - latest is 1.1.0 (MM is using 0.9.9)
* jquery-smooth-scroll - latest is 2.2.0 (MM is using 1.7.2)

mmistakes:
Some of those scripts are breaking changes if updated. Out of scope for this PR.

ohadschn:
Sure, was just wondering.
Should I open an issue for package update?
If so, one issue or issue per package?

_includes/head.html
mmistakes: Sure. Keeping things consistent with the rest of the "external URL" checks in the theme. You may be over thinking things a tad, simple is better.

mmistakes:
Don't worry about it. I have it covered. Some of these I may drop or replace with lighter alternatives. Now that scripts can be added easily I don't need to bundle everything and can leave it up to the user to augment if they want.

ohadschn:
Cool! Could you throw me a hint on which of these I should be OK with upgrading in my override?

_includes/head.html
ohadschn: Overthinking is my middle name 😅

mmistakes:
Not sure. You'll have to test.

If memory serves Magnific Popup had some CSS changes that I didn't want to mess with so I didn't bother updating.

Greedy nav had something I needed to override to fix a styling inconsistency.

The others I haven't looked at in a while.

ohadschn:
Understood. Thanks again for this. I guess I'll just follow the versions in the theme for now (simple enough to diff and update when a new release is out).

ohadschn:
Just a thought, maybe you could include something like `_main-only.min.js` (a minified version of _main.js` only) which could be used in the custom footer scripts approach rather than the "development-flavor" `_main.js `?

PR #1245, opened 2017-09-12
Add main-noplugins.min.js

ohadschn:
For custom footer script consumption per https://github.com/mmistakes/minimal-mistakes/pull/1241#issuecomment-328695872.


ohadschn:
I added `main-noplugins.min.js` to source control because that appears to be the convention, but I think ideally generated files shouldn't reside in the source control. I know very little about Ruby, but [according to the docs](http://guides.rubygems.org/command-reference/#gem-build), wouldn't the RakeFile be the most reasonable place to put the `uglifyjs` command?

mmistakes:
I'm not in love with any of these. The name `-noplugins` is misleading.

I really don't think this needs to live in the theme as it's not something most users will need. It's adding extra complexity which I'm not a fan of.

mmistakes:
Since all of the source files for the scripts are included, there is nothing stopping you from doing what you need to them. If that means running `uglify` to build just a minified version of `_main.js` without all the plugins included that should be doable.

If you're using a CI it's even easier since to my knowledge they allow you to run scripts at various points in the build process.

The reality is you've outgrown a pre-built theme for the masses and you either need to jump through hoops of this sort or better yet, build your own theme. That's exactly what I did years ago when I first started with Jekyll.

ohadschn:
Feel free to name it anything you want, I'm not in love with the name either :) As long as I have a file I can efficiently include in my footer script I'll be happy.

Seeing as the new footer script feature will prevent the inclusion of the `main.min.js`, and seeing as how users might want to use jQuery etc. in their custom scripts which would necessitate their separate inclusion, this would be a perfect fit.

IMHO literally less than one line of code that creates some files that as you said most people won't care about is not much of an added complexity...

ohadschn:
Yeah I suppose you are right about adding it to the CI, should be pretty trivial to add something like:
```bash
uglifyjs "$(bundle show minimal-mistakes-jekyll)/assets/js/_main.js" -c -m \
         -o assets/js/main-unbundled.min.js
``` 
(hope you like the new name)

However I decided to stick with the full `main.min.js` for now so I don't have to worry about adding/removing/changing plugins on upgrades.

PR #1255, opened 2017-09-16
Adding alt attribute for teaser images

mtlynch:
Without an alt attribute, HTML validators like HTML-Proofer will consider the
generated HTML to be incorrect. Also adds a strip filter to the title because
otherwise titles have a trailing newline.

### DIFF
--- a/_includes/archive-single.html
+++ b/_includes/archive-single.html
@@ -5,7 +5,7 @@
 {% endif %}
 
 {% if post.id %}
-  {% assign title = post.title | markdownify | remove: "<p>" | remove: "</p>" %}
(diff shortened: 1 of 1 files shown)


mmistakes:
This is somewhat subjective. Get why you'd want an `alt` description, but in it's current state it offers little value to screen readers here. Just adding the post's title to the image is redundant to the post title on the page already. If you turned on a screen reader or assistive technologies it would sound silly hearing it read the post title, then post title again as it hits the image.

Ideally it would describe the image, which does benefit screen reader users.

mtlynch:
Ah, that's true. Closing.

PR #1257, opened 2017-09-16
Added github: [metadata] necessary for running e.g. htmlproofer

lsolesen:


mmistakes:
Don't think this needs to be in the core theme as it's specific to someone using htmlproofer. I'd rather not set defaults that most people don't need.

PR #824, opened 2017-02-08
Changing translation in pt-br

pierryangelo:
There are some issues in the brazilian portuguese translation:
```  
less_than                  : "meno que" 
```
the word "meno" doesn't exist in the language, the correct translation is "menos"
```
  minute_read                : "minutos de leitura" 
```
once we can one or more minutes of reading, the plural of the word "minuto" is optional, so we can use "(s)" after the word "minuto" to reflect this

```
 related_label              : "Você Talvez Goste Também" 
```
this change is for better reading, the previous translation is not good

I changed the capitalization of the subsequent words for lowercase to have default capitalization in all translations.
You can see the results of this translation in my personal blog, [pierryangelo.github.io](pierryangelo.github.io)

Thanks.

PR #1275, opened 2017-09-25
Improve Brazilian Portuguese translation

lucascaton:


mmistakes:
Is this an improvement over what was done in #824? I'm not familiar with the language to know which reads better. There have been several PR's for this same change and just trying to figure out if this is just a "preference" or if one version is grammatically better.

mmistakes:
Trying to avoid accepting PR's were people keep changing the string to read how they think it should, which seems to be the case with all of the previous PR's.

lucascaton:
Hi @mmistakes, thanks for having a look at this.

The PR you mentioned is pretty much the same, so I'll close mine.
However, that was merged into `development` branch, is it going to be merged into `master` soon?

Thank you.

mmistakes:
It should have been merged into`master`. Will have to investigate why it wasn't. 

If you had time there are a [few text strings](https://github.com/mmistakes/minimal-mistakes/blob/master/_data/ui-text.yml#L256-L266) that are in need of translating if you had time to submit a PR.

lucascaton:
@mmistakes sure thing! https://github.com/mmistakes/minimal-mistakes/pull/1278

PR #1304, opened 2017-10-14
Polish locale

kbl:
Adding polish locale to ui-text.yml

### DIFF
--- a/_data/ui-text.yml
+++ b/_data/ui-text.yml
@@ -883,6 +883,49 @@ da: &DEFAULT_DA
 da-DK:
   <<: *DEFAULT_DA
 
+# Polish
+# ------
+pl: &DEFAULT_PL
+  page                       : "Strona"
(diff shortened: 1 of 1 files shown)


mmistakes:
Thanks for the PR!

kbl:
NP,  I should rather say - thank you for the theme ;) 

Unfortunately my translation isn't "perfect". In Polish there is a difference between "next" used with pages and posts. The same is true for reading time, depending on number "minutes" could have 3 different forms (https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/pl.yml#L106) ;) 

I don't know what is your intention there. Would you like rather to keep it simple/concise or support all language quirks.

mmistakes:
Simple and concise is fine. Since the data file is editable my intention is just to provide a starting point and if someone wants to tailor the text to suit their needs they can.

PR #1319, opened 2017-10-26
Fix google plus link

sumibi-yakitori:


mmistakes:
Can you give some detail on why this fix is needed? Google+ URLs with the `+` in them still exist and are valid. Example: <https://plus.google.com/+MichaelRoseDesign>.

sumibi-yakitori:
Hi

my google plus profile
https://plus.google.com/116211923005103602552

But this url is invalid
https://plus.google.com/+116211923005103602552

sumibi-yakitori:
Hi

my google plus profile
https://plus.google.com/116211923005103602552

But this url is invalid
https://plus.google.com/+116211923005103602552

mmistakes:
That makes sense then. I believe the `+` is for those who have set a vanity URL.

sumibi-yakitori:
I now know that google plus can make vanity url.
But my google plus account does not seem to be able to make vanity url.

You may reject this pull request.

mmistakes:
I've accepted the PR and will merge it in when I get a chance.

It's probably better to omit the `+` as you have to allow anyone with a profile to use it.

sumibi-yakitori:
Thank you.

PR #1320, opened 2017-10-26
Fix steam profile url

sumibi-yakitori:


mmistakes:
Can you give some examples or documentation on why this URL changed? I don't want to merge it in if there's a possibility people have `https://steamcommunity.com/id/` profile URLs as changing `id` ~> `profiles` would break their site.

sumibi-yakitori:
I'm sorry.
I was misunderstanding.

PR #1356, opened 2017-11-16
Fix lunar-en.js layout (Uncaught SyntaxError)

LouisMT:
Without layout set to null a HTML template will be rendered around the JavaScript, causing an error while loading the file in the browser.

```
lunr-en.js:1 Uncaught SyntaxError: Unexpected token <
```


mmistakes:
Thanks @LouisMT

PR #1357, opened 2017-11-16
Update social-share.html

DannieBGoode:
This way when clicking on the share on a social networks buttons, a share popup will open instead of forcing the share in the same window.


mmistakes:
What's the experience like on mobile? I see you have hardcoded widths in there at 500px. Will that force the main container to push out and mess up the layout?

mmistakes:
Thanks.... looks good!

PR #1363, opened 2017-11-21
Remove comments in dev environment

### LINKED ISSUE #1359: Disable analytics in development
- [X] This is a feature request.

---

## Environment informations

- **Minimal Mistakes version:** 4.7.1
- **Jekyll version:** 3.5.2
- **GitHub Pages hosted:** yes
- **Operating system:** macOS 10.10.5

---

## Expected behavior
I realised that [my blog](https://github.com/maaz93/maaz93.github.io)'s page visits were being tracked even in development. This is adding wrong entries into the site analytics. I would expect that the theme has a way to disable analytics in development.

Just to solve the issue quickly, I added `if (window.location.hostname !== "localhost") { ... }` in `google-universal.html`.

But, after [searching a bit](http://david.elbe.me/jekyll/2015/06/08/jekyll-different-settings-production-github-and-development.html), I found that we can use
```
{% if jekyll.environment == 'production' %}
  I'm in production!
{% else %} 
  I'm in development!
{% endif %}
```
I would add this change in `analytics.html`.

If you think this is a valid feature request, I wouldn't mind submitting a PR for this.

maaz93:
Fixes #1359

### DIFF
--- a/_layouts/single.html
+++ b/_layouts/single.html
@@ -63,7 +63,7 @@ layout: default
       {% include post_pagination.html %}
     </div>
 
-    {% if site.comments.provider and page.comments %}
+    {% if jekyll.environment == 'production' and site.comments.provider and page.comments %}
       {% include comments.html %}
     {% endif %}
   </article>


mmistakes:
Thanks @maaz93 for both of these PRs.

PR #1368, opened 2017-11-27
Fix button class

Stelios3g:
Hey @mmistakes,when i was trying to work with buttons,i found out that the button class needs 2 "--" instead of 1.Keep up the perfect job with the theme,we love it!


mmistakes:
Good catch, thanks!

PR #1369, opened 2017-11-27
Add option to search in specific collections

nickgarlis:
This is an enchancement for search results. With this you can turn **on/off** search for specific collections with `search: true`/ `search: false`. In this [example](http://vparticles.com/minimal-mistakes/search) search[ is disabled](https://github.com/nickgarlis/minimal-mistakes/blob/153c7195916f1b4cd20cb11e35c2caa877e40e40/docs/_config.yml#L235) for recipies. If you try searching for "chocolate" it will not come up with any results, however if you try searching for "oatmeal"
it will come up with the [post](https://github.com/nickgarlis/minimal-mistakes/blob/153c7195916f1b4cd20cb11e35c2caa877e40e40/docs/_recipes/oatmeal-cookies.md) in which search is enabled.


mmistakes:
To avoid having to set `search: true` in every document (or as a front matter default), what if the default behavior is everything is included in the search index unless it has `search: false`?

Then you can do this instead:

```
{% unless doc.search == false %}
...
{% endunless %}
```

nickgarlis:
Good idea, that would make more sense.

PR #1380, opened 2017-12-03
Tooltips for masthead links

### LINKED ISSUE #1377: Support tooltips for links in masthead
- [x] This is a feature request.

---

Sometimes the text displayed in the masthead links are not descriptive enough. If we could have tooltips for the links, we could give more details about what the link represents. For example, "Tags" can have a tooltip "Posts grouped by tags" as a tooltip.

I think this can be supported by adding an extra field in the _navigation.yml called `tooltip` and use it in the `masthead.html` like this

```
<li class="masthead__menu-item"><a href="{{ domain }}{{ link.url }}" title="{{ link.tooltip | default = link.title }}">{{ link.title }}</a></li>
```

If it seems like a valid request, I would love to submit a PR for this.

maaz93:
Fixes #1377

### DIFF
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -10,7 +10,9 @@
             {% else %}
               {% assign domain = site.url | append: site.baseurl %}
             {% endif %}
-            <li class="masthead__menu-item"><a href="{{ domain }}{{ link.url }}">{{ link.title }}</a></li>
+            <li class="masthead__menu-item">
+              <a href="{{ domain }}{{ link.url }}" {% if link.description %} title="{{ link.description }}" {% endif %}>{{ link.title }}</a>
+            </li>
           {% endfor %}
         </ul>
         <button type="button">
--- a/docs/_docs/07-navigation.md
+++ b/docs/_docs/07-navigation.md
@@ -34,6 +34,8 @@ Which will give you a responsive masthead similar to this:
 
 ![priority plus masthead animation]({{ "/assets/images/mm-priority-plus-masthead.gif" | absolute_url }})
 
+Optionally, you can add a `description` key per title in the `main` key. This `description` will show up like a tooltip, when the user hovers over the link on a desktop browser.
+
 **ProTip:** Put the most important links first so they're always visible and not hidden behind the **menu toggle**.
 {: .notice--info}
 


mmistakes:
Shouldn't this be something more like?

```html
<li class="masthead__menu-item">
  <a href="{{ domain }}{{ link.url }}" {% if link.description %}title="{{ link.description }}"{% endif %}>{{ link.title }}</a>
</li>
```

What you have is somewhat redundant and doesn't really produce HTML that would give the link more context since you're just duping `link.title` for the anchor `name` and `title` attribute`.

For example say you have this:

```yaml
main:
  - title: "My Masthead Link"
    url: http://whatever.com
```

You're going to get this...

```html
<li class="masthead__menu-item" title="My Masthead Link">
  <a href="http://whatever.com">My Masthead Link</a>
</li>
```

Which is kind of pointless since it doesn't add any additional context to the link.

https://www.searchenginejournal.com/how-to-use-link-title-attribute-correctly/

If you use my code:

```yaml
main:
  - title: "My Masthead Link"
    description: "An awesome website about whatever"
    url: http://whatever.com
```

You'd get this:

```html
<li class="masthead__menu-item">
  <a href="http://whatever.com" title="An awesome website about whatever">My Masthead Link</a>
</li>
```

Which I think is better for the user.

maaz93:
I added the title attribute on the wrong tag. I meant to add it to the anchor tag actually. Will change it.

I suggested about using a separate key `tooltip` in the `_navigation.yml` in the issue, but you [suggested](https://github.com/mmistakes/minimal-mistakes/issues/1377#issuecomment-348487622) that it won't add value. I think I misunderstood maybe? Do you think a key called `description` would make more sense? I'd be happy to add it if so 😄

mmistakes:
I was referring to how you were setting a default value for `title`. That is unnecessary as title is already the link name. 

`description` seems more semantic to me. The hover effect isn't really a "tooltip", just something browser's do. Not to mention it means nothing for mobile users since you can hover a link.

maaz93:
Fair enough. I've updated my commit. I've updated the docs as well.

mmistakes:
Looks good, thanks @maaz93.

Issue #1331, opened 2017-11-04
Linking Author avatar and name to site root


shredtechular:
- [X] This is a question about using the theme.
- [X] This is a feature request.
- [ ] I believe this to be a bug with the theme.
  - [ ] I have updated all gems with `bundle update`.
  - [ ] I have tested locally with `bundle exec jekyll build`.

## Feature Request Question

In your docs, you mentioned when in doubt on contributing to ask first, so this is just a question if you would want to consider a feature pull request for making the author avatar and name in the sidebar nav linkable -- for example, I have a customer interested in making those link back to the site root for users.

This might fall into the customization bucket that we should do this in our child of your theme, but wasn't sure if maybe you thought this might be desired by a number of your other users to consider a pull request for it?

Thanks!

mmistakes:
I think this might fall in the customization bucket. I'm not sure if this could be done "clean" so that it is easy to turn off/on for those who don't want it.

What makes it complicated as not everyone who'd want the author avatar and name to link back somewhere, will want it to go to the site's root. In the cases where blogs support [multiple authors](https://mmistakes.github.io/minimal-mistakes/layout-author-override/) this would be problematic.

This is why I've included the [website `uri` key](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/author-profile.html#L37-L43) for the author sidebar, to provide a way to link back to the author's site.

PR #1386, opened 2017-12-07
Feature: Adding linkback functionality Author avatar & Name

shredtechular:
This pull request is an attempt at addressing #1331.  This is the simplest/cleanest way that I can think to add this functionality.  If this is too messy, I completely understand and we can close this and I can simply do it locally, but again, I always like to keep customization as minimal as possible to more easily take advantage of future updates.

At any rate, I think this solution does solve the concern noted in #1331 around easily turning it on/off and allowing it for one, some, or many authors depending on the site's structure.

Basically, the code will check for an `author.linkback` yml key and if it is set, then it will add the link around a given author's avatar and name.  If it is not set, then the link will not be generated.

If this is something you'd consider adding in, I can certainly update the docs where applicable, but wanted to see if you'd even consider this before doing that?

### DIFF
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -10,29 +10,27 @@
     <div class="author__avatar">
       {% if author.avatar contains "://" %}
         {% assign author_src = author.avatar %}
-        {% assign author_class = "" %}
       {% else %}
         {% assign author_src = author.avatar | absolute_url %}
-        {% assign author_class = "author__avatar" %}
       {% endif %}
 
-      {% if author.linkback %}
-        {% if author.linkback contains "://" %}
-          {% assign author_link = author.linkback %}
+      {% if author.home %}
+        {% if author.home contains "://" %}
+          {% assign author_link = author.home %}
         {% else %}
-          {% assign author_link = author.linkback | absolute_url %}
+          {% assign author_link = author.home | absolute_url %}
         {% endif %}
         <a href="{{ author_link }}">
-          <img src="{{ author_src }}" class="{{ author_class }}" alt="{{ author.name }}" itemprop="image">
+          <img src="{{ author_src }}" alt="{{ author.name }}" itemprop="image">
         </a>
       {% else %}
-        <img src="{{ author_src }}" class="{{ author_class }}" alt="{{ author.name }}" itemprop="image">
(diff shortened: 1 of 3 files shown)


_includes/author-profile.html
```diff
@@ -9,15 +9,34 @@
   {% if author.avatar %}
     <div class="author__avatar">
       {% if author.avatar contains "://" %}
-        <img src="{{ author.avatar }}" alt="{{ author.name }}" itemprop="image">
+        {% assign author_src = author.avatar %}
+        {% assign author_class = "" %}
       {% else %}
-        <img src="{{ author.avatar | absolute_url }}" class="author__avatar" alt="{{ author.name }}" itemprop="image">
+        {% assign author_src = author.avatar | absolute_url %}
+        {% assign author_class = "author__avatar" %}
```
mmistakes: I think you can safely remove the `author_class` stuff. Looking back at my original code `class="author__avatar"` on the `` element was left in mistakenly. It's applied on the wrapper `div` and not needed.

mmistakes:
This looks pretty clean to me so I'm good with merging in with some minor adjustments.

Left a note above about the `author__avatar` class that was inconsistently applied to the bio image. I think I left that in by mistake, but it's not needed anymore.

The only thing giving me pause is the naming of `linkback`. Not sure if there is something better that is more clear or semantic. So if anyone has other suggestions I'm all ears. If nothing better comes in then `link_back` would follow the style of other `site` variables.

And yes, feel free to add this to the [Configuration docs](https://mmistakes.github.io/minimal-mistakes/docs/configuration/#site-author) and a line in `_config.yml` similar to the others.

shredtechular:
Thanks for the feedback!

I didn't realize that about `author__avatar` so I removed that in my update 890cf66.  I also updated the Configuration docs adding a line for the author site variable.

As far as the name of what variable to use there, I'm also open to suggestions on what might be the most clear.  Instead of `link_back` I changed it to simply `home` with the reasoning that folks that use this feature would probably be wanting to link to either the site's "home" page or an author "home" or "index" page.  Do you like that better or does that still not describe it adequately?

mmistakes:
`home` sounds good to me. Thanks for making the change.

PR #1387, opened 2017-12-07
Feature: Social Sharing Extensibility Include

shredtechular:
Similar to commit 2215ee5 that added easy customization extensibility for author links in the sidebar, this pull request does the same thing but for the Social Sharing section at the bottom of posts.

My specific use case was to add a "Discuss on" section underneath the theme's already defined "Social Sharing" section where users could click the link to discuss blog posts on specific social media sites.  So, it turned out looking like:

<kbd>
screen shot 2017-12-06 at 9 20 10 pm
</kbd>

###

I directly edited the `_includes/social-share.html` file in order to accomplish this, however, I thought it might be helpful to have something similar to the existing `includes/author-profile-custom-links.html` but for this section.  I think the added benefit here would give some additional flexibility for extending other social media sharing buttons here, even though it's clearly not difficult to edit that file either.  Anyways, wanted to share this if you thought useful to add in...

### DIFF
--- a/_includes/social-share-custom-links.html
+++ b/_includes/social-share-custom-links.html
@@ -1,0 +1,3 @@
+<!--
+	<a href="http://link-to-whatever-social-network.com?url={{ page.url | absolute_url | url_encode }}" class="btn btn--social-network" onclick="window.open(this.href, 'window', 'left=20,top=20,width=500,height=500,toolbar=1,resizable=0'); return false;" title="{{ site.data.ui-text[site.locale].share_on_label | default: 'Share on' }} Your Custom Social Network"><i class="fa fa-fw fa-social-network" aria-hidden="true"></i><span> Your Custom Social Network</span></a>
+-->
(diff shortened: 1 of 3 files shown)


mmistakes:
I'm torn with this one. I like the ability to add another "hook" for dropping in custom HTML/code into the template. But since you basically could add the same code to original `social-sharing.html` include, there's not much benefit.

Maybe another approach would be to pull out `social-share-custom-links.html` out of `social-share.html` and place it directly below as `social-share-custom.html`. This way if someone wants to replace the default sharing links with something else (native sharing buttons for example) they could.

The way you have it now that wouldn't happen, since they'd still get the default share links... then anything custom. Instead of completely replacing them. Plus they're wrapped in the `page.share` conditional to, which couldn't be altered.

shredtechular:
Hmm, I see what you are saying and I like the idea to 

> pull social-share-custom-links.html out of social-share.html and place it directly below it, as social-share-custom.html

for purposes of completely replacing the default social sharing buttons if desired... But then I think it does go back to what you were saying if they wanted to do that, why not just modify the `social-sharing.html` file, because any parent theme updates to that file wouldn't really matter anyway if those buttons were completely replaced.  So, that's why I was thinking this as more of an extension to the default than a replacement...

But, maybe this isn't the best solution and we should close it out.  Maybe instead of having those "default" buttons hardcoded into the `social-share.html` include file, that we use a yml data file/config to populate those buttons more dynamically and allow the end user to configure those buttons within yml?  Does that make sense and thoughts on if that would be better/preferable?

And if you'd rather not do either, I can easily close this out and live without it too!  Again, this was just something I was thinking of but can certainly get around it locally.

mmistakes:
I like the idea of using YAML data to generate the social links, but I haven't found a performant way of doing it.

To make it extensible you'd have to loop over the object of links (similar to the masthead links), and that starts to slow builds down... especially on large sites with tons of posts. This would be great for the sidebar author links too since I wouldn't have to maintain the Liquid spaghetti of if/elses.

And yeah I think this is going down the path of over complicating the theme so I'll go ahead and close it. The include for social links is almost never touched, so it's fairly safe for users to override/customize. 

Thanks for taking the time to discuss this and work on some solutions.

PR #1388, opened 2017-12-07
Update gems

### LINKED ISSUE #1384: Can I bump jekyll version?
- [x] I believe this to be a bug with the theme.

---

Since I'm using MRI 2.4+, the dependency fails because of old gems are incompatible with Ruby 2.4 in old `jekyll`. But I notice `github-pages` has bumped to `jekyll` 2.6 which should solve the problem. Mind a PR?

fantasticfears:
Fix #1384


mmistakes:
Looks good. Thanks @fantasticfears

PR #1411, opened 2017-12-19
Add Japanese translation of ui-text

sankichi92:
I didn't add the translations of "less than" and "minute read" for the following reasons.

- `read_time` does not work in Japanese documents.
- The Japanese translations of "less than" and "minute read" are "未満" and "分" respectively, but the word order is different like "1分未満."

Also, I didn't add the translations of "feed," "powered by," "website" and "email" because the alphabet was more natural.

---

I like this theme very much.  Thank you!


mmistakes:
Thanks!

PR #1433, opened 2018-01-01
Make excerpts optional in header overlay

dvhart:
Some posts and some header images don't lend themselves well to
displaying the excerpt in the overlay. Make this optional by
introducting a new boolean variable:

page.header.show_overlay_excerpt

Set it to default to true so existing users are unaffected.

This can be enabled globally for a site by changing the default to false
in the local _config.yml, or per page by specifying the value in front
matter.

Signed-off-by: Darren Hart <darren@dvhart.com>

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -252,3 +252,5 @@ defaults:
       comments: # true
       share: true
       related: true
+      header:
+        show_overlay_excerpt: true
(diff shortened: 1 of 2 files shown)


dvhart:
I was planning on doing a doc change at the same time, but it appears this should go in the gh-pages branch, so should be a separate PR.

dvhart:
Gah, found the docs dir - closing and will re-open a single PR with both change and doc.

PR #1434, opened 2018-01-01
Document show_overlay_excerpt variable

dvhart:
Add show_overlay_excerpt variable to the Header Overlay field
documentation table.

Signed-off-by: Darren Hart <darren@dvhart.com>

### DIFF
--- a/_docs/09-layouts.md
+++ b/_docs/09-layouts.md
@@ -250,13 +250,14 @@ header:
 
 To overlay text on top of a header image you have a few more options:
 
-| Name               | Description | Default |
-| ----               | ----------- | ------- |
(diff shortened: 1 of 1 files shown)


dvhart:
It wasn't clear to me if the gh-pages-3.1.6 was the right place to do this given the current theme version at 4.8.1, but it seemed like the a reasonable place to start.

dvhart:
Gah, found the docs dir - closing and will re-open a single PR with both change and doc.

mmistakes:
Yeah ignore that GH branch, it’s just there for archiving purposes.  All PRs should be based off `master`. All documentation files are under /docs.

PR #1435, opened 2018-01-01
Introduce show_overlay_excerpt and document

dvhart:
These two patches introduce a new feature allowing the per-page override of whether or not to display the excerpt in the overlay text, and document its usage.

### DIFF
--- a/docs/_docs/10-layouts.md
+++ b/docs/_docs/10-layouts.md
@@ -339,13 +339,14 @@ header:
 
 To overlay text on top of a header image you have a few more options:
 
-| Name               | Description | Default |
-| ----               | ----------- | ------- |
-| **overlay_image**  | Header image you'd like to overlay. Same rules as `header.image` from above. | |
-| **overlay_filter** | Color/opacity to overlay on top of the header image eg: `0.5` or `rgba(255, 0, 0, 0.5)`. |
-| **excerpt**        | Auto-generated page excerpt is added to the overlay text or can be overridden. | |
-| **cta_label**      | Call to action button text label. | `more_label` in UI Text data file |
-| **cta_url**        | Call to action button URL. | |
+| Name                     | Description | Default |
+| ----                     | ----------- | ------- |
+| **overlay_image**        | Header image you'd like to overlay. Same rules as `header.image` from above. | |
+| **overlay_filter**       | Color/opacity to overlay on top of the header image eg: `0.5` or `rgba(255, 0, 0, 0.5)`. |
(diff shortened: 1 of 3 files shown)


docs/_docs/10-layouts.md
```diff
@@ -339,13 +339,14 @@ header:
...
-| **overlay_image**  | Header image you'd like to overlay. Same rules as `header.image` from above. | |
-| **overlay_filter** | Color/opacity to overlay on top of the header image eg: `0.5` or `rgba(255, 0, 0, 0.5)`. |
-| **excerpt**        | Auto-generated page excerpt is added to the overlay text or can be overridden. | |
-| **cta_label**      | Call to action button text label. | `more_label` in UI Text data file |
-| **cta_url**        | Call to action button URL. | |
+| Name                     | Description | Default |
+| ----                     | ----------- | ------- |
+| **overlay_image**        | Header image you'd like to overlay. Same rules as `header.image` from above. | |
+| **overlay_filter**       | Color/opacity to overlay on top of the header image eg: `0.5` or `rgba(255, 0, 0, 0.5)`. |
+| **show_overlay_excerpt** | Display excerpt in the overlay test | true |
```
dvhart: And there's a typo: test should be text in the new line. I will resubmit after you've had a chance to review and roll in any changes you want me to make (like variable name for example).

mmistakes:
My only issue with this is the user has to explicitly set it as `true` to get the old behavior. So anyone who upgrades and doesn’t read the changelog will have no idea why their excerpts are gone.

Can this be done in a way were excerpts are shown always unless set to `false`? Perhaps this would work:

```liquid
{% if page.header.show_overlay_excerpt != false and page.excerpt %}
```

dvhart:
OK, this is a misunderstanding on my part regarding default values of variables. Does the _config.yml in this repository have no affect on default values? I set page.header.show_overlay_excerpt to true thinking that was how this is done. Clearly I'm brand spanking to new to jekyll, ruby, and gems. I'll test with your suggestion and resubmit.

mmistakes:
Default values work in the way you have them, but the config isn’t part of the theme. Users will need to add those values to their config file, and in my experience most will have no clue that is something they need to do.

Doing it the way I suggested makes the upgrade process easier. Existing users keep,the excerpts without breaking current functionality, and if they want to disable them they can add ‘false‘ globally to their config or page specific in their YAML front matter.

dvhart:
Agreed. That behavior was my intention. Testing the fix now.

Issue #1429, opened 2017-12-31
Split excerpt into archive_excerpt and overlay_excerpt


dvhart:
- [ ] This is a question about using the theme.
- [x] This is a feature request.
- [ ] I believe this to be a bug with the theme.
  - [ ] I have updated all gems with `bundle update`.
  - [ ] I have tested locally with `bundle exec jekyll build`.

---

## Environment informations

- **Minimal Mistakes version:** 4.8.1+
- **GitHub Pages hosted:** yes

---

## Expected behavior

The "excerpt" is used in at least 2 unique ways:

1. archive summary (archive-single)
  a) archive grid (e.g. related)
  b) recent posts (e.g. home)
2. header overlay (page__hero)

At least in my case, I'm looking for different things for 1 and 2 above, and was looking for a way to kill the excerpt from the header overlay, while keeping it for the archive summary. I didn't find a way, so I ultimately just commented it out of page__hero.

https://dvhart.github.io/2015/04/05/wannabee-it-guy-kid-friendly-networking.html

![screen shot 2017-12-31 at 11 03 31 am](https://user-images.githubusercontent.com/36253/34463731-4f308950-ee1a-11e7-96bf-50631a23fda5.png)

I think it would be useful to split excerpt into overlay_excerpt and archive_excerpt, both of which default to excerpt, but can be independently overridden. Existing users wouldn't need to change anything as overriding excerpt would apply to both new fields. In my case I would set 'overlay_excerpt: ""' in defaults in my _config.yml, and leave archive_excerpt.

Another option might be to create a boolean for overlay_excerpt which just toggles if the excerpt is displayed in in the overlay header. This would be less flexible, but probably less invasive.

dvhart:
The site wide boolean solution seems pretty minimally invasive, just the following to page__hero.html:

-      {% if page.excerpt %}
+      {% if page.excerpt and site.overlay_excerpt %}

And adding overlay_excerpt: (true|false) to _config.yml

dvhart:
If the above is acceptable, I can prepare a PR and update the documentation accordingly at the same time.

dvhart:
One last thought for now. page.header.overlay_excerpt as a boolean is even better. More flexibility, no more invasive.

mmistakes:
So to clarify your looking for a way to disable excerpts when using an image overlay? And not a way to specify different text for overlay excerpt and archive excerpts?

[...]

PR #1436, opened 2018-01-01
v2: Introduce show_overlay_excerpt and document 

dvhart:
These two patches introduce a new feature (#1429) allowing the per-page override of whether or not to display the excerpt in the overlay text, and document its usage.

v2:
  - Correct typo in documentat (test -> text)
  - Explicitly test for "!= false" to avoid changing default behavior
  - Remove _config.yml changes as they are no longer needed to maintain default behavior


mmistakes:
This looks good @dvhart! Thanks for making the changes and pushing this through.

PR #1449, opened 2018-01-04
 Minimize JSON data

nickgarlis:
I was thinking of maybe moving the last function ( where the search takes place and the results are displayed ) into a new file too but I am not currently sure if it's going to stay the same for every language that I add. 
 
[Live Preview](http://vparticles.com/minimal-mistakes/)


mmistakes:
Looking good. We might have to think about truncating the excerpts size, but looking at the the `lunr-store.js` for the MM demo site it's not too bad.

I'm showing around 148 kB for an index with all of the body content. Which is nice since those searches for `"search"` actually produce results now. Guess if someone has a 1,000 post site it might be too big, but otherwise I'm fine with leaving it as is for now.

_includes/scripts.html
```diff
@@ -19,8 +19,9 @@
   {% else %}
     {% assign lang = "en" %}
   {% endcase %}
-  <script src="{{ '/assets/js/lunr.min.js' | absolute_url }}"></script>
-  <script src="{{ '/assets/js/lunr-' | append: lang | append: '.js' | absolute_url }}"></script>
+  <script src="{{ '/assets/js/lunr/lunr.min.js' | absolute_url }}"></script>
```
mmistakes: Probably worth adding `defer` to each of these three scripts to give priority to the other page content loading first.

eg. `<script defer src="{{ '/assets/js/lunr/lunr-...js' | absolute_url }}"></script>`

nickgarlis: Hmm great idea! We could also decide whether to truncate or not based on the number of posts there are.

mmistakes: I like that idea of truncating based on the number of posts. Not sure what a good baseline should be. Guess it really depends on the the size of `{{ content }}`.

In the case of the MM demo site, the documentation pages are pretty big and full of words.

Maybe we default to truncating (can probably up it to well over 50 words) and then have an override setting in `_config.yml` to index the full post. If the user wants to take the potential hit to page load perf they can.

mmistakes: Maybe something like:

```yaml
#_config.yml
search_full_content: true # true, false (default)
```

```html
<!-- lunr-store.js -->
{% if site.search_full_content == true %}
  {% assign excerpt = doc.content | strip_html | strip_newlines | jsonify %}
{% else %}
  {% assign excerpt = doc.content | strip_html | truncatewords: 20 | jsonify %}
{% endif %}
```

nickgarlis: What if there is a safe zone parameter which gets activated when there are over a thousand posts and its default size is set to 50 but can also be modified to be lower or higher ? If that's too complicated then we could have some pre-set parameters like: safe which would be somewhere in the 50s and super-safe which would be 20.

mmistakes: I’m not convinced post count is the way to go, way too unreliable. I could have 2,000 posts with a sentence of text each or have a couple hundred with epic word counts going over 10,000 and get a huge ‘lunar-store.js‘ because it falls below the threshold.

I think truncating the content to around 50 words by default, with a config override to turn that off would be sensible. If someone wants to get more granular with the amount of words truncated they override the theme file.

nickgarlis: Yes it really depends on the individual. The switch is great option too since it's more straight forward if the user has too much content they simply turn off the search_full_content parameter.

PR #1471, opened 2018-01-14
Use title as alt on the image

lsolesen:


mmistakes:
Closing as this is a duplicate of #1255.

That PR was closed because adding the post's `title` as an `alt` description adds zero value. An alt description should describe the image and not simply repeat the post's title which is already outputted on the page.

Issue #1473, opened 2018-01-14
Add optional ability to display date of post in the "archive-single.html"


BR0kEN-:
### Synopsis

I think it would be good to have an optional ability to display the date of a post (from `post.date`) in the `archive-single.html` template.

### Technical approach

To not break the operability of the current implementation I propose to introduce the `date_show` argument to the `include` statement. That way those who interested in displaying the date would be able to optionally step in and turn the feature on. Also, alongside with the first argument, it's reasonable to allow configuring the date format, introducing the `date_format` argument that defaults to the `%B %-d, %Y`.

mmistakes:
See #1474 for context as to why it was closed.

PR #1474, opened 2018-01-14
#1473: Add optional ability to display date of post in the "archive-single.html"

BR0kEN-:
https://github.com/mmistakes/minimal-mistakes/issues/1473


mmistakes:
Thanks for the PR but I'm trying to balance making everything toggleable. The theme is already gotten out of hand because of I haven't been as diligent with not merging in these sorts of edits.

In my opinion the date in an archive is of little value. But of course if you disagree that's what theme overrides are for by adding your own modified layouts/includes to your local build.

PR #1476, opened 2018-01-16
Fix atom feed in _config.yml 

snupt:
Custom path doesn't work for atom. Read README file here for more information https://github.com/jekyll/jekyll-feed

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -49,7 +49,7 @@ staticman:
 reCaptcha:
   siteKey                :
   secret                 :
-atom_feed:
+feed:
   path                   : # blank (default) uses feed.xml
 search                   : # true, false (default)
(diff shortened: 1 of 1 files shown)


mmistakes:
This PR is incomplete and breaks functionality. The theme uses `atom_feed.path` [in many locations](https://github.com/mmistakes/minimal-mistakes/search?utf8=%E2%9C%93&q=atom_feed&type=). Simply changing it to `feed.path` in _config.yml isn't enough.

Closing as I don't think this change is necessary. You still should set `atom_feed.path` for a custom feed to be used in the theme's footer, and also set `feed.path` to the same if you're hosting on GitHub Pages and want it to handle that redirect for you.

PR #1481, opened 2018-01-16
Add 'hide_site_title' flag to pages

lucascaton:
Hi there!

There are some pages that I don't want to have a title suffix, ie.: instead of `My Page - My Site`, I'd like to only have `My Page`.
Does it make sense? Please let me know if there's a better/easier way to do it :)

Cheers.

### DIFF
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -5,7 +5,11 @@
 {%- assign seo_url = seo_url | default: site.github.url -%}
 
 {%- if page.title -%}
-  {%- assign seo_title = page.title | append: " " | append: site.title_separator | append: " " | append: site.title -%}
+  {%- if page.hide_site_title -%}
+    {%- assign seo_title = page.title -%}
+  {%- else -%}
+    {%- assign seo_title = page.title | append: " " | append: site.title_separator | append: " " | append: site.title -%}
+  {%- endif -%}
 {%- endif -%}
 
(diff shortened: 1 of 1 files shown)


mmistakes:
What problem does this PR propose to solve?

lucascaton:
@mmistakes sorry - I should've been more clear :)

There are some pages that I don't want to have a title suffix, ie.: instead of `My Page - My Site`, I'd like to only have `My Page`. Does it make sense?

mmistakes:
Makes sense. Seems like a niche use case and not something that would benefit most users of the theme, especially with it adding more complexity. 

Closing as this is best left as something you override/customize locally yourself.

lucascaton:
@mmistakes I customized it locally already, I only submitted the PR in case it could help anyone else.

The only problem is that `_includes/seo.html` is a 166 LOC file and it's hard to remember to keep it in sync with its original file.

lucascaton:
hey @JoshBeardJB - you can add it to any page, as long as it includes your HTML template.

For example, you can add it to the following `_pages/about.md` file:

```md
---
layout: single
title: About
permalink: /about/
hide_site_title: true
---

Your page content here.
```

Hope it helps :)

JoshBeardJB:
Yes! Okay, I got it now @lucascaton - Thank you so much for taking the time to respond!

PR #1485, opened 2018-01-17
Fix jsonify name in seo.html application/ld+json script

d00616:
When the name includes &quot; then the resulting schema is invalid. This is a fix.

### DIFF
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -137,7 +137,7 @@
     {
       "@context" : "http://schema.org",
       "@type" : "{% if site.social.type %}{{ site.social.type }}{% else %}Person{% endif %}",
-      "name" : "{{ site.social.name | default: site.name }}",
+      "name" : {{ site.social.name | default: site.name | jsonify }},
       "url" : {{ seo_url | jsonify }},
       "sameAs" : {{ site.social.links | jsonify }}
     }


mmistakes:
Thanks for the PR.

d00616:
To reproduce, set the "name:" field to a string "This is a &quot;Test&quot;" and validate the result via https://search.google.com/structured-data/testing-tool

Without this fix seo.html produces something like:
```
<script type="application/ld+json">
    {
      "@context" : "http://schema.org",
      "@type" : "Organization",
      "name" : "This is a "Test"",
      "url" : "http://1.2.3.4:4000",
      "sameAs" : null
    }
  </script>
```

d00616:
The result with jsonify is invalid

```
"name" : ""This is a \"Test\"""
```

mmistakes:
@d00616 You have to remove the quotes around the variable. To keep things consistent I'd prefer to use `jsonify` like the others.

This would be the change:

```html
<script type="application/ld+json">
{
  "@context": "http://schema.org",
   "@type": "{% if site.social.type %}{{ site.social.type }}{% else %}Person{% endif %}",
    "name": {{ site.social.name | default: site.name | jsonify }},
    "url": {{ seo_url | jsonify }},
   "sameAs": {{ site.social.links | jsonify }}
}
</script>
```

Which produces this valid JSON:

```html
<script type="application/ld+json">
{
  "@context": "http://schema.org",
  "@type": "Person",
  "name": "This is a \"Test\"",
  "url": null,
  "sameAs": null
}
</script>
```

d00616:
It's changed.

d00616:
Thank you.

PR #1487, opened 2018-01-18
fix a typo in 04-upgrading.md

googlr:
It's a trivial typo that I found when reading your documents.

In "minimal-mistakes/docs/_docs/04-upgrading.md", it seems a verb is missing in "To check which version you are currently using, view the source of your built site and you should something similar to:
".

Thank you for the amazing Jekll theme. 👍


mmistakes:
Thanks!

PR #1501, opened 2018-01-26
Accessibility improvements

jasonday:
Included:

- HTML5 landmarks for the header
- Improved labeling & interactions on Search for keyboard only and screenreader users
- Improved interactions & context for Hamburger menu

### DIFF
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -1,4 +1,4 @@
-<header role="banner" class="masthead">
+<header class="masthead">
   <div class="masthead__inner-wrap">
     <div class="masthead__menu">
       <nav id="site-nav" class="greedy-nav">


_includes/masthead.html
```diff
@@ -1,4 +1,4 @@
-<div class="masthead">
+<header role="banner" class="masthead">
```
mmistakes: I think `role="banner"` is redundant. According to the W3 using `<header>` when its a descendant of the body automatically gets `role="banner"`.

> The HTML5 header element defines a banner landmark when its context is the body element.
>
> The HTML5 header element is not considered a banner landmark when it is descendant of any of following elements: `article`, `aside`, `main`, `nav`, `section`
>
> _ref: [Banner Landmark HTML5 techniques](https://www.w3.org/TR/wai-aria-practices/examples/landmarks/banner.html)_

jasonday:
Change made.

mmistakes:
Thanks @jasonday for the quick change. I want to give everything else a look over and make sure none of script functionality breaks when they're squashed together with `npm run build:js`.

After that should be good to merge in.

coliff:
A related suggestion; I think the search input should be `type="search"` rather than `type="text"`

mmistakes:
@jasonday Would you be able to look at the merge conflicts? Looks like the search form include is out of sync with the latest since I added support for Algolia.

It's been moved to [`_includes/search/search_form.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/search/search_form.html). There's some `case` logic in there but shouldn't be too bad to migrate your changes.

One other thing I'd like to add would be to replace text strings with `site.data.ui-text` strings, like I've done with the rest of the theme. This will allow it to be localized in the `_data/ui-text.yml`.

For example the site search visually hidden label:

```html
<label class="visually-hidden" for="search">Site Search</label>
```

Would become:

```html
<label class="visually-hidden" for="search">{{ site.data.ui-text[site.locale].site_search_label | default: 'Site Search' }}</label>
```

And then in `_data/ui-text.yml` a line under the `en` key would need to be added:

```yaml
site_search_label: "Site Search"
```

and a blank value for all the other languages so they could be localized in the future:

```yaml
# es, fr, tr, pt, pt-BR, it, zh, de, ne, ko, ru, li, gr, sv, nl, id, vi, da, po, ja
site_search_label:
```

jasonday:
Will do. but I probably won't get to this until the weekend.

mmistakes:
@jasonday No rush. Whenever you have the time.

PR #1511, opened 2018-01-31
Rectify German UI text

TpmKranz:
In German the polite second person pronouns („Sie“, „Ihr“) are always capitalized and compound words never have whitespace in between.

Also, more appropriate Translations.


mmistakes:
Thanks @TpmKranz

Issue #1480, opened 2018-01-16
QUestion: how to make content area wider


ddd-workshop:
- [x ] This is a question about using the theme.
- [ ] This is a feature request.
- [ ] I believe this to be a bug with the theme.

---

How can I make the article size wider (I'm using single layout) on larger screens so that the content is not limited to only to the narrow part of the whole page as on the attached screenshot: 
![width](https://user-images.githubusercontent.com/16007327/34991020-70b00064-fac8-11e7-8f96-2212e4fa4e8c.png)

mmistakes:
Please search through the closed issues. I've replied to this with examples and actual code more times than I can count.

mmistakes:
cc/ #1373, #1265

Issue #1373, opened 2017-11-29
Help with getting wider post content


m1evankaiser:
- [x] This is a question about using the theme.
- [ ] This is a feature request.
- [ ] I believe this to be a bug with the theme.
  - [ ] I have updated all gems with `bundle update`.
  - [ ] I have tested locally with `bundle exec jekyll build`.

---

## Environment informations

- **Minimal Mistakes version:** 4.6.0
- **Jekyll version:** 3.6.2
- **GitHub Pages hosted:** yes
- **Operating system:** MacOS Sierra

---

## Expected behavior

I'd like to know how to adjust the width of the content such that the code blocks seen on [this page](http://devblog.members1st.org/dotnet-core-rest-api-in-about-30-minutes/) show more content without scrolling.

Ironically, if I shrink the width of the browser window until it gets to the large and medium breakpoints, you can see a lot more of the text without scrolling. However, the x-large view is both a narrower column of text and bigger font, resulting in a hard to read post. 

I found in `_variables.scss`, that it has a series of `$right-sidebar-width` variables, however, lowering those, which did help increase the content width, also seemed to negatively affect the left-sidebar (making it more squished) and I don't want to affect the left-sidebar, as it is fine the way it is.

## Steps to reproduce the behavior

Example page: http://devblog.members1st.org/dotnet-core-rest-api-in-about-30-minutes/
Repo: https://github.com/Members1stFCU/Members1stFCU.github.io

Steps to reproduce:
- Cloned the theme
- Customized `_config.yml`, `_data/navigation.yml`, and some other non-style related changes
- Added some posts
- Tested locally
- Pushed up to GitHub Pages

mmistakes:
Have a search through the closed issues, I've covered this one a few times before.

You need to alter the CSS some. If you look through `_sass/_page.scss` you'll see I'm applying some negative margins on the right. That's "faking" some space to make sure an optional right sidebar fits. If you want the main content to expand into that space, removing the margins will do it.

```scss
.page {
  width: calc(100% - #{$right-sidebar-width-narrow});
}
```

mmistakes:
https://github.com/mmistakes/minimal-mistakes/issues/1265

m1evankaiser:
Thanks that is much better! I saw a mention about removing the sidebar padding, but thought that meant changing the sidebar width values in _variables.scss.

Issue #1265, opened 2017-09-19
how to adjust the width of the second column?


taochenshh:
Hi, thanks for the amazing theme.

How can I change the width of the second column of this two-column style theme. For example, I wanna enlarge the width of the blog list so that each line can show more words? 

Thank you for your help.

mmistakes:
You'll need to adjust the theme's CSS on [`.archive`](https://github.com/mmistakes/minimal-mistakes/blob/master/_sass/minimal-mistakes/_archive.scss#L5-L17) and [`.page`.](https://github.com/mmistakes/minimal-mistakes/blob/master/_sass/minimal-mistakes/_page.scss#L23-L34) If you remove the right padding it should give you the effect you want.

![image](https://user-images.githubusercontent.com/1376749/30591249-ffffa092-9d0f-11e7-8023-e362a609ba16.png)

Related: #623, #384, #74

taochenshh:
Thanks.

amosshi:
See #2093

Issue #384, opened 2016-06-30
The effect of varying the width of Browser Window


healiseu:
- [X] This is a question about using the theme.
- [X] I have updated all gems with `bundle update`.
- [X] I have tested locally with `bundle exec jekyll build`.
- [X] I have not modified assets in such a way that will affect the width of any element

---

## Environment Information

- **Minimal Mistakes version:** 3.2.10
- **`github-pages` or `jekyll` gem version:** 3.1.6
- **Operating system:** Ubuntu

---
Hi Michael,
I have studied a bit a [similar issue that was posted almost two years ago](https://github.com/mmistakes/minimal-mistakes/issues/74) and then discussed again recently. In one of your comments there you mentioned that "**you are working on a complete rewrite of the theme which should make customizing things a bit easier**". This is version 3.2.10 of the theme, therefore I decided to open a new issue and ask a similar question.

![image](https://cloud.githubusercontent.com/assets/14239698/16487866/00008884-3ed6-11e6-8850-fd5d2e68df09.png)

This is a 1275x684 image captured from my Browser. My desktop background can be seen at the vertical edges of the image. You can measure the left and right margin of the images grid using a screen ruler and find that it is 2x120pixels = 240pixels. The total length of the browser window is about 1240pixes wide and therefore subtracting the margins we are left with a content area width of about 1000pixels.

##### My question is what are the theme files I should modify and where exactly to increase the content area width, i.e. decrease the left and right margin of the browser window ?

To make things a bit more interesting, let us put in the game the sidebar.

![image](https://cloud.githubusercontent.com/assets/14239698/16488400/052984a2-3ed9-11e6-84b9-3cdb9d7b1812.png)

The width of the image captured from my Browser is the same and the width of the browser window has not been modified. If we measure again the right margin from the edge of the browser window, we find that it is 300 pixels ! I am not sure what is the width of the sidebar, it will be nice to help me put a background color, but If I measure from the point the social media images start I get 120pixels. This is the same margin I had without the side bar. 

##### This indicates that somehow when you add the sidebar, the width of the content area is shrinked because the right margin has been increased ! How is it possible to fix the right margin, e.g. 120 pixels in this example ? 

You may also consider that If you experiment with even larger sizes of a browser window, e.g. 1900px then you get a right margin of about 520px wide and a left margin of about 300px !

Thanks in advance for any effort to solve that problem and to answer my question.

[...]

PR #1516, opened 2018-02-02
Add `wide` variant to single layout

peterahrens:
This pull request adds a new layout called "queen." It is a wider version of "single" (yes like the mattress sizes). It puts the right sidebar on top (if you asked for a right sidebar) so that you can get that sweet sweet wide display you've been craving for your programming blog post (80 characters or bust!), without cramping your style in your more literary blog posts. I don't really know if this layout is really worth merging, but I thought I would put it here so somebody might copy what I did instead of bugging mmistakes about widening the margins again.


peterahrens:
honestly, I am just going to close this and some diligent github user might read it later.

mmistakes:
The elegant approach here is to utilize the `classes` front matter instead of adding a new `layout`.

For a page you'd want the wide variant of you'd do something like:

```yaml
---
layout: single
classes: wide
---
```

Which would put a `wide` class on the [`<body>`](https://github.com/mmistakes/minimal-mistakes/blob/master/_layouts/default.html#L17) element. You could then and similar CSS to what you did but using `.wide` to override the page and sidebar classes.

```scss
.wide {
  
  .page {
    // override right padding and width's to make wide
  }

  .sidebar__right {
    // override negative margin, width, positioning, etc.
  }

  // any other overrides needed for a wide `.page`
}
```

mmistakes:
If you want to take a stab at this I'll review it and see about getting it merged in.

mmistakes:
Looks good @peterahrens, thanks for making the change and resubmitting.

peterahrens:
Thanks for the tips @mmistakes!

peterahrens:
Hopefully this helps with issues like #1480, #1373, #1265, #384, #136, and #74.

ohadschn:
Do you guys have a live demo for this?

mmistakes:
https://mmistakes.github.io/minimal-mistakes/markup-text-readability-wide-page/

ohadschn:
Nice!
Apologies for my laziness, I should have checked your sample post list - comprehensive as always :)

dongdongbh:
I add classes: wide to post, but it didn't work? I didn't use toc, and use same ccs file as the master branch

PR #1566, opened 2018-03-07
Fix typos

TristianK3604:
I was looking at the `read-time.html` file to see how it worked and noticed that it said `elsif` rather than `elseif`.


mmistakes:
This isn't a typo. Liquid uses `elsif` and not `elseif`. Changing to that will break the logic.

ref: https://help.shopify.com/themes/liquid/tags/control-flow-tags#else-elsif

PR #1608, opened 2018-03-30
Refactor author profile

lsolesen:
This small change will make it possible to include author information manually other places also in your site - like in the bottom of e.g. http://historie.vih.dk/om/


lsolesen:
@mmistakes I still think this pull request is a good idea. Any reason why it just ended up stale? - and not as a won't fix?

PR #1624, opened 2018-04-18
My new branch

jspidersf:
Deleting unnecessary files.


jspidersf:
deleting unnecessary files.

PR #1628, opened 2018-04-20
Adding a Mailchimp Powered Pop-up working 'on-click' (non-intrusive)

pateskinasy:
Script to make Mailchimp pop-up work 'on click' (Mailchimp's provided code pops up automatically in your screen - not a fan-). 

1) You should replace your own baseURL in the script (from pop-up code Mailchimp provides you)
2) You should replace your own uuid in the script (from pop-up code Mailchimp provides you)
3) You should replace your own lid in the script (from pop-up code Mailchimp provides you)

### DIFF
--- a/_includes/newsletter.html
+++ b/_includes/newsletter.html
@@ -1,0 +1,18 @@
+<script type="text/javascript" src="//downloads.mailchimp.com/js/signup-forms/popup/embed.js" data-dojo-config="usePlainJson: true, isDebug: false">
+</script>
+<script>
+function showMailingPopUp() {
+  require(
+    ["mojo/signup-forms/Loader"],
+    function(L) {
+      L.start({"baseUrl":"yourbaseurl","uuid":"youruuid","lid":"yourlid"})
+    }
+  );
+
+  document.cookie = "MCPopupClosed=;path=/;expires=Thu, 01 Jan 1970 00:00:00 UTC;";
+  document.cookie = "MCPopupSubscribed=;path=/;expires=Thu, 01 Jan 1970 00:00:00 UTC;";
+}
+
+document.getElementById("open-popup").onclick = function() {showMailingPopUp()};
+
+</script>
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -14,6 +14,7 @@
               <a href="{{ domain }}{{ link.url }}" {% if link.description %}title="{{ link.description }}"{% endif %}>{{ link.title }}</a>
             </li>
           {% endfor %}
+          <a href="#" id="open-popup" onclick ="showMailingPopUp(); return false;">Join Newsletter</a>
         </ul>
         {% if site.search == true %}
         <button class="search__toggle" type="button">
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -37,6 +37,7 @@
     </div>
 
     {% include scripts.html %}
+    {% include newsletter.html %}
 
   </body>
 </html>


_includes/newsletter.html
```diff
@@ -0,0 +1,18 @@
+<script type="text/javascript" src="//downloads.mailchimp.com/js/signup-forms/popup/embed.js" data-dojo-config="usePlainJson: true, isDebug: false">
```
coliff: I suggest removing `type="text/javascript"` (not needed in HTML5) and making the requested URL specifically over HTTPS (src="https://downloads....), its potentially safer.

coliff:
Adding this would mean all the code for this would be added to every page of every installation of this theme - even for users who don't use or need Mailchimp sign integration. I think it'd be better if this was an opt-in option

mmistakes:
Agreed @coliff. I'm not sure if this is even something worth merging into the theme. Is there a big enough need for it?

@pateskinasy If it were to be merged it would need to be made more flexible... namely making `yourbaseurl`, `youruuid`, and `yourlid` variables in the script something theme user's could set from `_config.yml`. And as @coliff pointed out, would need to be opt-in. Meaning you'd probably want to wrap `{% include newsletter.html %}` with some sort of Liquid condition:

```html
{% if site.mc_baseurl and site.mc_uuid and site.mc_lid %}
  {% include newsletter.html %}
{% endif %}
```

With better named variables obvioulsy :wink:

_includes/newsletter.html
pateskinasy: Thanks for the feedback @coliff . That specific line came straight from Mailchimp's 'pop-up code'. I've customized the rest of the code to make it work on-click and not automatically. 
Independently, true, bringing HTTPS in and 'type' won't hurt. More comments from me below.

pateskinasy:
Thanks for the feedback @coliff and @mmistakes .

Jumping into @mmistakes' question on the need for it:

The main reason I've decided to bring this improvement suggestion to the project is because I'm using Minimal Mistakes for a blog. Given that 'by nature', a stand-alone blog doesn't have social features (like Medium's Follows/Applauds, etc), managing a list of "subscribers" you can push new posts and newsletters to, may help getting bloggers' new posts discovered and more of your content seen.

Mailchimp's code allows automatic pop-ups to collect emails out of the box (you'll find it in your Mailchimp dashboard). But they are annoying 🤷 as they pop-up automatically to every visitor to your blog 😞. The code above makes the pop-up work on-click.

Should you guys think this can be a good addition, I can give the improvements with liquid and variables a go on the weekend. Let me know.

mmistakes:
Sounds good @pateskinasy and thanks for the PR.

I'll leave this open. Let's see if users chime in if this would be useful. I'm trying to avoid bloating the theme with features that aren't used by more than 80% of its users. As something like this can easily be added to `scripts.html` or `footer/custom.html` includes already provided... I don't see much reason to merge it in.

But let's see if it's something a bunch of people want and can convince me to add it.

maxime-michel:
If not a popup, I would vote for the theme to at least propose a generic Mailchimp form. Some sort of notice box with a title, some text, and the ability to add fields through YAML. This is something I use across websites and the code duplication makes me sad. What do you guys think? I can submit a PR.

maxime-michel:
This is what I currently run: https://github.com/maxime-michel/asg/commit/b248524e8426a662b2f7ae9c067faa7a94f12dd1. With some checks on Liquid variables and a little help on the design side, I believe it could be a good addition.

mailchimp

pateskinasy:
Well, I have on my to-do list for improvements on my blog to add a simple one line input field (no box) form for newsletter in my sidebar, in my footer, and in as a 50px hello bar top of screen. So, I’m with you in 👌🏼

PR #1664, opened 2018-05-10
fix broken recipe collection and make pets work

jamiesmith:
Nice theme.  I was experimenting with it and noticed that the link to recipes was broken, and that the link to pets wasn't complete.


mmistakes:
Looks good, thanks!

PR #1670, opened 2018-05-12
allow external link _target

jamiesmith:
I like to have external links have the option to open in a new tab.  This allows users to specify that.

### DIFF
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -11,7 +11,7 @@
               {% assign domain = site.url | append: site.baseurl %}
             {% endif %}
             <li class="masthead__menu-item">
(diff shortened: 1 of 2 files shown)


mmistakes:
Thanks for the PR, but I'm not a fan of forcing `target="_blank"` on the user for all of the reasons cited here: https://css-tricks.com/use-target_blank/

jamiesmith:
ok

Sciss:
You can add another liquid thing: https://github.com/Sciss/minimal-mistakes/blob/homepage2/_plugins/extlink_tag.rb

It works like this:

```
{% extlink MIBsaal http://www.musikerinitiative-bremen.de/ %}
```

PR #1696, opened 2018-06-03
DRY `tags_max` calculation in _layouts/tags.html

e28eta:
It was being done twice, and it doesn't seem to need to be.

### DIFF
--- a/_layouts/tags.html
+++ b/_layouts/tags.html
@@ -4,13 +4,14 @@ layout: archive
 
 {{ content }}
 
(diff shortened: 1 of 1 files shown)


e28eta:
Again, thanks for all the hard work you do on this theme! As I was working on #1697, I saw this potential opportunity to clean it up.

I wasn't sure if you'd want both of these changes, so for simplicity I split it up.

mmistakes:
Thanks for this. Looks good!

PR #1697, opened 2018-06-03
Add minimum_tag_count param to _layouts/tags.html

e28eta:
Adds a `page` parameter `minimum_tag_count`. Any tags with *fewer* than `minimum_tag_count` (no bounds checking, less than 1 is like 1) will be omitted from the tags page.

My tags (currently) outnumber the posts, and I felt it was really noisy going to the Tags page and seeing a long list of tags with only one post. So, adding this parameter allows adding `minimum_tag_count: 2` to the front matter of the Tags page, and it hides any tag with just 1 post.

It *is* semi-broken because the tag on the post links to the Tags page, which then doesn't have that tag.

Thanks for all the awesome work on this theme, and I won't mind if you don't think this is right for the project (or if there's a better way to achieve this). I had the idea, tried it out, and thought it might be worth sharing. I don't know if this is useful enough to merge, or if I'm in the minority wanting this feature.

### DIFF
--- a/_layouts/tags.html
+++ b/_layouts/tags.html
@@ -1,8 +1,11 @@
 ---
 layout: archive
+minimum_tag_count: 1
 ---
 
 {{ content }}
+
+{% assign tags_min = page.minimum_tag_count %}
 
 <ul class="taxonomy__index">
   {% assign tags_max = 0 %}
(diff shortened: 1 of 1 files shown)


mmistakes:
I can see the utility of this, but as you mention it breaks the anchor references to tags that might be omitted on the "tag archive page".

This seems like the perfect use case for doing a local override of the layout if it's something you need.

Thanks for taking the time to submit the PR. I'll give it the **enhancement** tag so others can potentially find it in the closed issues and use if they'd like.

PR #1701, opened 2018-06-09
Include a default title_separator

Matir:
The current code prevented any separator being included when site.title_separator was unset.


mmistakes:
Looks good, thanks!

PR #1728, opened 2018-07-03
Changes to list of files that can be deleted

hello-w0r1d:
Removing the minimal-mistakes-jekyll.gemspec file breaks the ability to run locally using "bundle exec jekyll serve".


mmistakes:
You shouldn't be using the gemspec file. That is for development only.
Instead it should be replaced with a [`Gemfile`](https://mmistakes.github.io/minimal-mistakes/docs/quick-start-guide/#migrating-to-gem-version) and depending on if you're self-hosting or using GitHub Pages you'll either add `gem "jekyll"` or `gem "github-pages"` to it.

mmistakes:
Also forking the theme is discouraged. If installed as a ruby gem or using `remote_theme` the issue with the `.gemspec` file doesn't exist.

PR #1741, opened 2018-07-10
Update french labels

flohw:
Here are some updates on french labels.

A lot of time since I used the theme. It's usage as a ruby gem is a lot easier for updates ! Thank you for that ! 😃


mmistakes:
Thanks as always!

PR #1751, opened 2018-07-19
trival_fix_of_doc

qudongfang:
if `minimal-mistakes-jekyll.gemspec` is deleted, `Gemfile` should be deleted too,
otherwise there will be an error when you run  `jekyll serve` 
such as "There was an error parsing `Gemfile`: There are no gemspecs at /tmp/minimal-mistakes-master. Bundler cannot continue."

### DIFF
--- a/docs/_docs/01-quick-start-guide.md
+++ b/docs/_docs/01-quick-start-guide.md
@@ -92,6 +92,7 @@ If you forked or downloaded the `minimal-mistakes-jekyll` repo you can safely re
 - `/test`
 - `CHANGELOG.md`
 - `minimal-mistakes-jekyll.gemspec`
+- `Gemfile`
 - `README.md`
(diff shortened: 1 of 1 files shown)


mmistakes:
`Gemfile` is needed. I guess the docs could be more descriptive and explain that the `Gemfile` in the root of the project needs to be updated as  in it's current state it's for development only.

[This section is where I explain that](https://mmistakes.github.io/minimal-mistakes/docs/quick-start-guide/#migrating-to-gem-version). I think the only issue is when people fork the theme. They really shouldn't be starting with the root files as they're incomplete and configured for building the theme gem. A better start is the files in [`/docs`](https://github.com/mmistakes/minimal-mistakes/tree/master/docs).

qudongfang:
OK, Thanks. I will close this pull request then.

PR #1756, opened 2018-07-22
Allow time to be shown with a header image overlay

bexelbie:
This fixes the apparent typo that required both the site and page read_time to be true before the time would be shown.


mmistakes:
Thanks for this.

`site.read_time` isn't a thing and was added in error. To put this in-line with [`_includes/archive-single.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/archive-single.html#L33-L35) you can go with just:

```
{% if page.read_time %}
  ...
{% endif %}
```

And remove `site.page_read_time or`

bexelbie:
@mmistakes PR updated.

PR #1786, opened 2018-08-11
Fix smooth scroll breaking back button #1767

### LINKED ISSUE #1767: Smooth scroll breaks back button
### Current Behavior

Smooth scroll is triggered when a link to a different section on the same page is clicked, e.g. when clicking a table of contents. After clicking a section on a table of contents, users often want to go back to the table of contents. However, users cannot do that because the URL is not appended with the hash (section ID). This can be demonstrated at https://kswedberg.github.io/jquery-smooth-scroll/demo/.

### Proposed Solution:

1. Disable smooth scroll; or
2. Enable hashchange feature of jQuery smooth scroll. There are two ways to do that:
    a. Using native hashchange DOM event: [demo](https://kswedberg.github.io/jquery-smooth-scroll/demo/hashchange.html), [source code](https://github.com/kswedberg/jquery-smooth-scroll/blob/master/demo/hashchange.html).
    b. Not using native hashchange DOM event: [demo](https://kswedberg.github.io/jquery-smooth-scroll/demo/bbq.html), [source code](https://github.com/kswedberg/jquery-smooth-scroll/blob/master/demo/bbq.html).

danieljl:
Fix #1767.


mmistakes:
Is there any way to make the URLs that appears in the browser bar match the ones on the page?

When testing this fix on the [`/test` site content](https://github.com/mmistakes/minimal-mistakes/tree/master/test) and a page with a table of contents. All of the URL's there are `/#hash`. When clicking on them the URL bar changes to `/#/hash`. And if you share those URLs they don't scroll down to the appropriate location.

Where as `/#hash` does.

image

image

mmistakes:
They scroll for me when clicking fine. What I'm concerned about is if those links get indexed or shared and someone clicks on it from an external source or different page, then it doesn't load at the proper heading. It just starts at the top of the page.

I'll take a look at the script and see if I can alter the regex to keep `/#hash` instead of replacing it with `/#/hash`.

danieljl:
> What I'm concerned about is if those links get indexed or shared and someone clicks on it from an external source or different page, then it doesn't load at the proper heading. It just starts at the top of the page.

How did you test it? The way I tested was by copying the link with a hash and pasting in a new tab (not in the current tab). And the loaded page started at the section the hash referred to.

mmistakes:
`/#/hash` doesn't work when JavaScript is disabled.
`/#hash` does work when JavaScript is disabled because it's a valid anchor due to the headline elements having `id="hash"` attributes.

PR #1843, opened 2018-09-20
Remove recommendation to add reCaptcha to staticman.yml

danamlewis:
Suggestion for docs - adding to staticman causes reCaptcha to not show in the comment form; only `_config.yml` needed modifying.

(Adding to staticman blocks the reCaptcha from showing in the site - or at least did in my [implementation](https://github.com/opening-pathways/website) of the template. Otherwise following these docs worked great for adding reCaptcha in.)

And - huge thank you for this site template *and* the fantastic documentation! I've highlighted it [here](http://openingpathways.org/everything-is-open-source) as one of the reason's we chose it for our project's blog.

### DIFF
--- a/docs/_docs/05-configuration.md
+++ b/docs/_docs/05-configuration.md
@@ -441,7 +441,7 @@ To skip this moderation step simply set `moderation: false`.
 To enable Google's reCAPTCHA to aid in spam detection you'll need to:
 
 1. Apply for [reCAPTCHA API](https://www.google.com/recaptcha) keys and register your site using the reCAPTCHA V2 type.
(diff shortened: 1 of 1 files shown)


mmistakes:
Strange. As per Staticman's docs they're needed in its `staticman.yml` config.

https://staticman.net/docs/configuration

danamlewis:
Ok. It's possible that I implemented weirdly, somehow! But, reCaptcha is working successfully in mine without being in staticman, just _config.

mmistakes:
Looks like you've [hardcoded the reCaptcha values](https://github.com/opening-pathways/website/blob/master/_includes/comments.html#L91) directly into the comments include. So it's not coming from `_config.yml` at all. Guessing there's more of that since your repo appears to be modified in a few places.

I'd be curious to hear what other's experiences are with configuring Staticman v2 with reCaptcha and if it deviates from what's in the docs.

Issue #1842, opened 2018-09-19
Public Staticman server is hitting GitHub API limits


maxime-michel:
I would like to raise awareness about recent issues me and others have experienced with Staticman.

The first is that it is becoming almost impossible to get an 'OK' back when adding the bot as a repository collaborator: https://github.com/eduardoboucas/staticman/issues/227.

The second is that some comments might possibly get lost due to the public instance (`api.staticman.net`) hitting GitHub API limits: https://github.com/eduardoboucas/staticman/issues/222

The solution to those problems is to allow people to define a custom URL for their own deployment of Staticman. But I also believe some awareness about the potential issues should be present in the documentation.

Another idea could be that the theme transparently uses its own deployment of Staticman. It would greatly reduce the simplicity if we had one instance to host N users, rather than N users paying for and setting up one instance individually. But somebody's got to handle it at the end of the day, and that would have to be well organized.

Any thoughts?

PS: this is an excellent theme and a great framework! I wish I had joined the party sooner.

mmistakes:
Thanks for bringing it up. Not sure on the best path forward.
I guess anyone who relies on the public instance probably needs a backup plan. Kind of comes with the territory when using "free" stuff.

maxime-michel:
Would you be OK with a PR that opens the endpoint URL to possible configuration?

mmistakes:
Yeah I'd be cool with that. Could probably do something like `{{ site.staticman.endpoint | default: 'https://api.staticman.net/v2/entry/' }}` to fallback to the public instance if one isn't specified in `_config.yml`.

PR #1845, opened 2018-09-20
Allow custom Staticman endpoints

maxime-michel:
As discussed in #1842.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -53,6 +53,7 @@ staticman:
       type               : # "date"
       options:
         format           : # "iso8601" (default), "timestamp-seconds", "timestamp-milliseconds"
+  endpoint               : # URL of your own deployment with trailing slash, will fallback to the public instance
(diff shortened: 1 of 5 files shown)


maxime-michel:
Sure. I saw you initially requested that but I wasn't sure it was what you actually meant. Then I saw an occurrence of `{{ site.comments.disqus.shortname }}`, which made me go for this approach. But sure, I don't mind.

mmistakes:
Sounds good.

In the grand scheme of things it probably doesn't matter. What you have was a good approach since it follows the other comment providers. If there wasn't already `site,staticman` your approach makes the most sense.

maxime-michel:
@mmistakes I don't mean to be spammy but did you get a notification for the changes I pushed right after your comment?

mmistakes:
No didn't get a notification. But I see it now.

PR #1847, opened 2018-09-23
Fix missing title with custom comments

diegoscl:
Added the <h4> tag to show the comments title when custom comment provider is selected.

### DIFF
--- a/_includes/comments.html
+++ b/_includes/comments.html
@@ -166,6 +166,7 @@
         {% endif %}
       </section>
     {% when "custom" %}
+      <h4 class="page__comments-title">{{ comments_label }}</h4>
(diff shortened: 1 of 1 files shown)


_includes/comments.html
```diff
@@ -166,6 +166,7 @@ <h4 class="page__comments-title">{{ site.data.ui-text[site.locale].comments_labe
         {% endif %}
       </section>
     {% when "custom" %}
+      <h4 class="page__comments-title">{{ comments_label }}</h4>
```
mmistakes: Variable is incorrect. Should be `{{ site.data.ui-text[site.locale].comments_label | default: "Leave a Comment" }}`

PR #1856, opened 2018-09-29
Open external footer links in new window

rmarquis:
This is a bug fix.

## Summary

The "Powered by" links in the footer area open in the same window by default. This commit changes the behaviour to open them in a new window.

## Context

This is particularly useful when hosting the project on GitHub pages or GitLab pages, and using a domain name that masks the URL in a frame, so the external links don't open under the same domain name.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -16,4 +16,4 @@
   </ul>
 </div>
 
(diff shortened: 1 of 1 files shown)


mmistakes:
Thanks for the PR, but I'm not a fan of forcing this behavior on the user.

See this excellent post on CSS Tricks for all the reasons why this is bad and not justified.

rmarquis:
Well, the author in the link you provide states this is very opinionated, but I understand this is more an issue of my own situation (DNS masking through frame). Anyway, not a big deal, I'll simply patch it locally.

Thanks for your time!

PR #1865, opened 2018-10-02
adding Mastodon to social profiles in config.yml

frankmeeuwsen:

## Summary

In the _config.yml I included the possibility to add your Mastodon profile in both the sidebar and footer. You can add your full Mastodon profile URL to the list.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -113,6 +113,9 @@ author:
     - label: "Website"
       icon: "fas fa-fw fa-link"
       # url: "https://your-website.com"
+    - label: "Mastodon"
+      icon: "fab fa-fw fa-mastodon"
+      # url: 
     - label: "Twitter"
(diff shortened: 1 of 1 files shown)


mmistakes:
Closing as the theme allows you add any link/icon to the author sidebar and footer via `author.links` and `footer.links` in _config.yml

Check the docs for more info and examples.

https://mmistakes.github.io/minimal-mistakes/docs/configuration/#site-author

frankmeeuwsen:
I know you can do this with the `author.links` and `footer.links`. But why would there be a couple of defaults already then, like Gitlab and Facebook? Why not add one of the newer platforms as well?  Because, you might as well remove the defaults from Twitter, Facebook, Github and Instagram as well since _these_ can be added as well with `author.links` and `footer.links`.

mmistakes:
Those are deperecated and will eventually be removed.

PR #1874, opened 2018-10-04
Cache "static" includes to improve build performance

mmistakes:

## Summary

Includes that are computational expensive and are called several times can potentially see a improvement in build time if cached. Includes that do not rely on the `page` context are prime candidates for caching e.g. the masthead, footer, scripts, etc.

Install [jekyll-include-cache](https://github.com/benbalter/jekyll-include-cache) and replace includes that fit this criteria using `{{ include_cached }}` tag.

[**`/test` build times (Windows 7):**](https://github.com/mmistakes/minimal-mistakes/tree/master/test)

|  | `{{ include }}` | `{{ include_cached }}` |
| --- | --- | --- |
| cold start | 11.442s | 10.633s |
| rebuild | 10.136s | 9.136s |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Enhancements
 
+- Cache "static" includes to improve build performance. [#1874](https://github.com/mmistakes/minimal-mistakes/pull/1874)
 - Make entire feature and archive items "clickable". [#1864](https://github.com/mmistakes/minimal-mistakes/pull/1864)
 - Allow custom Staticman endpoints. [#1842](https://github.com/mmistakes/minimal-mistakes/issues/1842)
 - Remove `type="text/css"` from Algolia script includes. [#1836](https://github.com/mmistakes/minimal-mistakes/pull/1836)
--- a/_config.yml
+++ b/_config.yml
@@ -226,6 +226,7 @@ plugins:
   - jekyll-gist
   - jekyll-feed
   - jemoji
+  - jekyll-include-cache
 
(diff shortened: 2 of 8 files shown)


bradvolen:
Getting google page build errors. Even after adding gem "jekyll-include-cache" to my Gemfile.

The page build failed for the `master` branch with the following error:

The tag `include_cached` on line 16 in `/_layouts/default.html` is not a recognized Liquid tag. For more information, see https://help.github.com/articles/page-build-failed-unknown-tag-error/.

mmistakes:
@bradvolen Thanks for pointing this out. I've added notes to the theme's documentation about it.

While GitHub Pages supports this plugin it looks like it's not turned on by default because it has the potentially to be used incorrectly. I guess this is why you need to manually add it to the `plugins` array of _config.yml

justinrummel:
Just curious, should lines 13-14 _layouts/default.html also have `include_cached`?

https://github.com/mmistakes/minimal-mistakes/blob/e704cd40d4c8629742f949ba012f2a64200fffc3/_layouts/default.html#L13-L14

mmistakes:
Head can't because it contains jekyll-seo-tag which pulls from the `page` object to populate all the SEO stuff. Using `include_cached` would break that.

**head/custom.html** might be OK, but I left it off of that as I have no idea what someone might be using it for. If they have `page.whatever` variables in there it would break.

Pretty much the only thing you can use `{% include_cached %}` for is if it's straight HTML or Markdown. Or only uses `site.whatever` variables.

PR #1884, opened 2018-10-08
Fix Lunr search index merging words

### LINKED ISSUE #1883: Lunr search index merging words across newlines
The lunr index in minimal-mistakes is fed using an excerpt from each page in /assets/js/lunr/lunr-store.js. The excerpt is passed through `strip_newlines` which removes any newlines in the excerpt, but does not replace those newlines with anything, resulting in merged words across newlines (ex: `foo\nbar` becomes `foobar`, but should become `foo bar`).

So our search index has many merged words.

I'm not sure where this should be fixed. One option would be to do something stupid in minimal-mistakes like `newline_to_br` followed by `replace: '<br>', ' '` rather than the `strip_newlines`, but probably the "right" solution is to modify `strip_newlines` to replace with spaces, or add an option or new function in Liquid to do that. It looks like Liquid's `replace` cannot easily to used on newlines directly (which is presumably why the `newline` specific function exist).

What do you think?

bmcgough:
This is a bug fix.

## Context

Fixes #1883

### DIFF
--- a/assets/js/lunr/lunr-store.js
+++ b/assets/js/lunr/lunr-store.js
@@ -18,7 +18,8 @@ var store = [
         "title": {{ doc.title | jsonify }},
         "excerpt":
           {%- if site.search_full_content == true -%}
-            {{ doc.content |
+            {{ doc.content | newline_to_br |
+              replace:"<br />", " " |
               replace:"</p>", " " |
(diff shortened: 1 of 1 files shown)


mmistakes:
Looks good. Thanks @bmcgough

PR #1885, opened 2018-10-08
Updated LUNR to 2.3.3

justinrummel:
## Summary

Updated `lunr.js` and `lunr.min.js` to current version [2.3.3](https://github.com/olivernn/lunr.js/blob/v2.3.3/lunr.js).  Local testing shows Console errors and results are correct.


mmistakes:
Looks good. Thanks @justinrummel

PR #1918, opened 2018-10-29
Added async tag

Kulbhushan-Chand:

## Summary

Optimizing page speed by asynchronously loading the java scripts.
With the addition of async tag, the browser will execute javascript as soon as it is available (asynchronously), instead of waiting for all of the JS resources to load and parsed. This results in improvement in page speed performance.

## Issue
https://github.com/mmistakes/minimal-mistakes/issues/1919#issue-375037559

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -5,11 +5,11 @@
     {% else %}
       {% capture script_path %}{{ script | relative_url }}{% endcapture %}
     {% endif %}
-    <script src="{{ script_path }}"></script>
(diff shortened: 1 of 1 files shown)


Kulbhushan-Chand:
I'm closing this pull request, since you are using similar attribute (`defer`) on `fontawesome.js`. 
Regarding use of `async `on `main.js`, I'm not sure if there is any other dependent scripts used elsewhere in the theme which might gets affected.

Issue #1920, opened 2018-10-29
Missing subresource integrity checking for Font Awesome JS


Kristaba:
It seems that the Font Awesome's javascript code is served by their CDN by default.
I am not too much used to web development, but I believe the [subresource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) checking should be used in such case.

https://github.com/mmistakes/minimal-mistakes/blob/f4cfc87b0a4a4f3150ab47cd4491ac46834db6a1/_includes/scripts.html#L12

The hash of each version of Font Awesome files is available in  [their website](https://fontawesome.com/how-to-use/on-the-web/setup/getting-started?using=svg-with-js).
So the current line should be replaced by something like:
```html
<script src="https://use.fontawesome.com/releases/v5.3.1/js/all.js" integrity="sha384-kW+oWsYx3YpxvjtZjFXqazFpA7UP/MbiY4jvs+RWZo2+N94PFZ36T6TFkc9O3qoB"></script>
```

mmistakes:
Care to submit a pull request with the `integrity` value added?

PR #1922, opened 2018-10-30
add integrity hash for Font Awesome script

Kristaba:
Proposed fix for #1920. 
As the script is loaded from an external CDN (use.fontawesome.com), its integrity must be ensured mainly for security reasons.
This hash must be updated together with the exact version of Font Awesome, as given on their website.

I have **not** added the `crossorigin="anonymous"` policy as proposed by Font Awesome as it is a quite different topic, with implications I don't fully understand.

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -9,7 +9,7 @@
   {% endfor %}
 {% else %}
   <script src="{{ '/assets/js/main.min.js' | relative_url }}"></script>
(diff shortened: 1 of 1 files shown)


DeepWalter:
It seems that only adding the `integrity` part will lead to font-awesome icons not showing up on Github pages. e.g. on [Minimal Mistakes homepage](https://mmistakes.github.io/minimal-mistakes/) footer, all the icons are missing.

Not sure why, but either add `crossorigin="anonymous"` or remove the `integrity` part will fix the problem.

mmistakes:
@DeepWalter `integrity` value was wrong, so the script didn't load properly. It's fixed now.

DeepWalter:
Thank you @mmistakes !

PR #1924, opened 2018-11-02
Add rel attribute to social media links in author profile and fotter

torrocus:

## Summary

Add rel attribute to social media links. 
Set value noopener for security reason. Otherwise you can get to opener information using `window.opener.document`. 
Set value nofollow to inform search engines that we don't want to count a link as a vote for that url. Social media have no problem getting users. They don't need dofollow links.
Set value norefferer to hide referrer information when the link is clicked. This is an option for more privacy in social media. They know so much.

## Context

It's not related to any GitHub issue.


mmistakes:
Thanks!

PR #1969, opened 2018-11-22
Changed schema.org to secure links (https)

Kulbhushan-Chand:

## Summary
Using secure links (https) for schema.org

## Context

It's not related to any GitHub issue.

### DIFF
--- a/_includes/breadcrumbs.html
+++ b/_includes/breadcrumbs.html
@@ -13,12 +13,12 @@
 {% endif %}
 
 <nav class="breadcrumbs">
-  <ol itemscope itemtype="http://schema.org/BreadcrumbList">
+  <ol itemscope itemtype="https://schema.org/BreadcrumbList">
     {% assign crumbs = page.url | split: '/' %}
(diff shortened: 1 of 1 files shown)


mmistakes:
Do you know if this is documented anywhere? All the schemes.org examples are `http://`

I do see they are serving the site over `https`. Just want to make sure this doesn’t break anything.

mmistakes:
Never mind, found the answer in their faqs.

> **Q: Should we write 'https://schema.org' or 'http://schema.org' in our markup?**
>
> There is a general trend towards using 'https' more widely, and you can already write 'https://schema.org' in your structured data. Over time we will migrate the schema.org site itself towards using https: as the default version of the site and our preferred form in examples. However 'http://schema.org' -based URLs in structured data markup will remain widely understood for the forseeable future and there should be no urgency about migrating existing data. This is a lengthy way of saying that both 'https://schema.org' and 'http://schema.org' are fine.

mmistakes:
@justinrummel Yes, that would be great.

PR #1976, opened 2018-11-25
Prevent current post from showing in the "You may also enjoy" (related

sestus:
posts) section

This is a bug fix.

## Summary

When I forked this repo and created my first post, I noticed that in the "You may also enjoy" section, this first post was include, which doesn't make much sense. I have added a check to skip the current post when populating the related_posts.


sestus:
Forgot to say thank you for your awesome theme and the extensive documentation :smile: . It's awesome!

mmistakes:
Looks good. Thanks for the kind words and the PR.

PR #1983, opened 2018-12-02
Google Custom Search does not work if not using Instant Search

efung:
fbafe58e40b1bd4a2628aef1eb5955ab57dcfaf4 renamed the function in the
scripts file, but did not rename the call sites.

This is a bug fix.

## Summary

Google Custom Search integration does not work if Instant Search is not enabled.

## Context

In my `_config.yml`, I had:

```
google:
  search_engine_id       : "my_id"
  instant_search         : # false (default), true
```

When I tried to execute a search, I got an error in the Javascript console about `executeQuery` not being defined.

As part of #1667, a function was renamed in fbafe58e40b1bd4a2628aef1eb5955ab57dcfaf4. However, the two call sites, in `_includes/search/search_form.html` and `_layouts/search.html`, were not renamed.


mmistakes:
Thanks @efung

PR #1907, opened 2018-10-23
Revert caching for comments and analytics

paulfioravanti:
This is a bug fix.

## Summary

Reverts caching for comments and analytics providers.

## Context

Fixes #1905. Although the issue only covered comments, I've uncached the `analytics.html` file as well since:

- the include that brings it in (`{% include scripts.html %}`) is now uncached
- it [uses a `page` variable](https://github.com/mmistakes/minimal-mistakes/blob/c8f4820d31c6b4c605ec96f74213df007c74c7cd/_includes/analytics.html#L1) that [can't be cached to work properly](https://github.com/mmistakes/minimal-mistakes/issues/1905#issuecomment-432040761)

PR #1985, opened 2018-12-03
Caching analytics and search includes

Kulbhushan-Chand:
## Summary

Caching analytics and search includes for build performance
You have also cached `analytics script include` on your personal website - [here](https://github.com/mmistakes/made-mistakes-jekyll/blob/f83304755ebb0efd58103bf3d7cf240205e61cf6/src/_includes/scripts.html#L2)

## Context
It is not related to any GitHub issue

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -24,5 +24,5 @@
   {%- endcase -%}
 {% endif %}
 
-{% include analytics.html %}
-{% include /comments-providers/scripts.html %}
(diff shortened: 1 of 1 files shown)


mmistakes:
These were previously cached, but were reverted as doing so caused issues. See https://github.com/mmistakes/minimal-mistakes/pull/1907

Kulbhushan-Chand:
Thanks for the link to issue because of which I've managed to identify a bug on my website.

PR #1988, opened 2018-12-05
Update video

gammell:
Fix comment tag.

### DIFF
--- a/_includes/video
+++ b/_includes/video
@@ -1,7 +1,7 @@
 {% capture video_id %}{{ include.id }}{% endcapture %}
 {% capture video_provider %}{{ include.provider }}{% endcapture %}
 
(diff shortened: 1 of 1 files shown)


gammell:
Sorry for the confusion, my change is to remove the slash, not add it.

mmistakes:
Gotcha, still don't think the slashes matter. You have a link to a repo where you're experiencing the issue.
Seems to be working fine for the demo site.

gammell:
Unfortunately, there's no repo at the moment. I'm in the process of working through my various bugs, and while I figured it was unlikely that that comment was causing general problems, I thought I'd flag it since it was nonstandard.

Thanks for the great theme by the way.

mmistakes:
OK. Closing for now. If you can reproduce and provide a repo I'm happy to look into it further.

PR #2004, opened 2018-12-20
Persian Translation of UI Elements

rouyeen:
**This is an enhancement or feature.**

## Summary

Adds Persian (Farsi) to the supported UI localizations.

Note: In Persian, the location where the "less_than" label should appear is different. To avoid unnecessary modifications, "less than" is simply ignored by replacing it with a single space. The only downside is that small reading times will be rounded up to 1 minute.

## Context

No related GitHub issues.


mmistakes:
Thanks for the PR @rouyeen

PR #2014, opened 2019-01-01
Accessibility: Add aria-label to the search button

SeanKilleen:

## Summary
I recently switched my site to use the latest minimal mistakes theme. I ran an audit and noticed that it scored 68 for accessibility.

I looked into it and of the issues is that the search toggle button does not have a discernable name: 

> screenshot of the chrome audit

When [looking into it](https://dequeuniversity.com/rules/axe/2.2/button-name?application=lighthouse) I saw it could be fixed by adding an `aria-label`, so I figured I'd contribute that back.

This PR adds an `aria-label` to the search toggle button. I'm open to any preferences on the specific wording to use.

### DIFF
--- a/_data/ui-text.yml
+++ b/_data/ui-text.yml
@@ -9,6 +9,7 @@ en: &DEFAULT_EN
   breadcrumb_home_label      : "Home"
   breadcrumb_separator       : "/"
   menu_label                 : "Toggle menu"
+  search_label               : "Toggle search"
   toc_label                  : "On this page"
   ext_link_label             : "Direct link"
   less_than                  : "less than"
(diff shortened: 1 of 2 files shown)


SeanKilleen:
@mmistakes The edits make sense 👍 I'll get onto them as soon as I get the chance, probably tonight.

SeanKilleen:
@mmistakes ok, changes made 👍 I think this is in line with what you're looking for.

SeanKilleen:
btw while I'm at this: do you have any desire for a shortcut & conversion for the i18n piece? E.g. to turn `{{ site.data.ui-text[site.locale].search_label | default: "Toggle Search" }}` into something like `{% i18n toggle_search %}`? If so, I could do a separate pr for that.

SeanKilleen:
@mmistakes I think I just accomplished it locally; I'll create an issue to discuss it further. 👍

mmistakes:
Everything looks great, thanks for the quick fixes.

PR #2016, opened 2019-01-02
[On Hold] i18n include

SeanKilleen:

## Summary
Allows the i18n text to be formatted as `{% include i18n item="itemname" %}`

Example, for `search_label`: `{% include i18n item="search_label" %}`


mmistakes:
I'll have to think on this some. First brush, I feel it over complicates things and doesn't really improve much.

It also has the drawback of not being able to specify default text since it relies on the existence of a `ui-text.yml` file. I've found that that file can go missing quite often when users install the theme since it's not "bundled" with it.

The current method can still pickup `default` text because it's an actual string in the the theme files that gets pulled in if there is no `_data/ui-text.yml` or a string for that key in the file.

SeanKilleen:
@mmistakes sure 👍  at this point, at least you have the example should you decide to proceed with it. 

No worries if you don't decide to accept it -- I totally understand and I was using it as a way to further explore liquid, which I've done 😄 . Definitely not worth it if it doesn't add value. I was thinking the value proposition was in less verbose template files & easier ability to add i18n text without lengthy strings. 

I'm going to mark this as "On hold" for now. Feel free to close if you decide against it, no hard feelings.

mmistakes:
Thanks @SeanKilleen. I'll keep it open in case others might find it useful, and let the stale bot close it in 30 days if there is no further activity.

SeanKilleen:
@mmistakes thanks! To clarify -- your intent is to close it by default even if I was to complete this PR, correct? That's fine; I just wouldn't want to complete the work if it didn't have a chance of being merged.

PR #2019, opened 2019-01-04
Clicking TOC links changes hash fragment

### LINKED ISSUE #1302: Anchor links in TOC don't change URL hash
- [X] This is a question about using the theme.
  - [X] I have updated all gems with `bundle update`.
  - [X] I have tested locally with `bundle exec jekyll build`.

---

## Environment informations

- **Minimal Mistakes version:** (4.6.0, 4.5.2)
- **Jekyll version:** 3.6.0
- **GitHub Pages hosted:** yes
- **Operating system:** Ubuntu 16.04

---

## Expected behavior

When I click an anchor link in the TOC generated with `{% include toc %}`, the active URL will change to `site/page/#anchor`.

This, so that when the page is reloaded, the content from the anchor is in focus.

- On <https://mmistakes.github.io/minimal-mistakes/docs/layouts/>, hover over any link in the TOC ("Included layouts"), target hover displays expected URL `site/page/#anchor`.
- Click the link, page scrolls to anchor
- URL is still `site/page/`(i.e., https://mmistakes.github.io/minimal-mistakes/docs/layouts/), not `site/page/#anchor`.

I assume that this is defined somewhere in the JavaScript, but as I'm not sure which technique is used I see myself unable to find the part where the location hash is removed. Also I'm not sure whether this is part of the theme or happens somewhere else?

edemaine:
This is a bug fix / enhancement to TOC behavior.

## Summary

Change URL's hash when clicking on a TOC item, and respond to hash changes (e.g. clicking back button).

## Context

This PR fixes #1302 (not sure why closed).  It is based on #1767, and indeed I based off of that.  It should fix all outstanding issues:
* No hash mangling (`#section`, not `#/section`).
* Use `history.pushState` to behave like clicking without bumping into default (nonsmooth) `location.hash = ...` behavior.

## Additional

I'd also be happy to add automatic hash changing behavior when scrolling (as on e.g. https://coffeescript.org/), if that's of interest, either in this PR or another.


mmistakes:
Great work @edemaine! I just tested this out and everything looks perfect to me.

Can you take care of the merge conflict (I think it's just the date in the banner copyright in main.min.js... you probably forked before I made that change)? Then I can get this merged in.

And yes, please open another PR with your proposal for automatic hash changing on scroll.

edemaine:
@mmistakes Thanks for the quick review! I rebased to master, and upon further testing, added a commit that switches from `hashchange` to `popstate` event handler.  This seems to behave better with the Forward button (wasn't scrolling before), on page load (scrolled inconsistently), and always scrolls in the right direction (before, it seemed to go to the top of the page and scroll down, even if using the back button ought to just scroll you up).  This should be ready to go now!

maxime-michel:
I just noticed a small issue with this where clicking on the TOC won't jump to the title when said title contains special characters. The generated HTML is:

    <li><a href="#rel%C3%A2chement-des-jambes">Relâchement des jambes</a></li>
    <li><a href="#bien-%C3%AAtre-int%C3%A9gral">Bien-être intégral</a></li>

But the JS tries to hit `#relâchement-des-jambes`, and obviously there's no match. Which of the two is preferable, though?

edemaine:
@maxime-michel Sorry for the delay. My reading of [RFC 3986](https://tools.ietf.org/html/rfc3986) defining URLs suggests that `#...%C3%AA` really should be equivalent to `#...ê`, i.e., the percent encodings should in fact be decoded.  Clearly something (presumably some JavaScript) is buggy, though.  Have you noticed whether this was specific to my changes, or perhaps a general smooth-scrolling bug?  I will investigate.

maxime-michel:
Thanks, I actually had never tried this before seeing this new feature. I just did some experimenting and realised that no matter whether the `href` attribute is encoded or not, it'll go to the correction location, as long as the `id` is being encoded. Which isn't the case currently. I'll submit a PR.

maxime-michel:
No problem. No, I'd never tried the shortcuts before I read this feature was introduced.

On a non-Jekyll vanilla HTML page, as long as the subtitle's `id` is encoded, both versions of the hash can access it. With the theme, though, first of all the title isn't encoded (but there isn't much that could be done about from the theme, I guess). But even when I override it to the encoded version, that still doesn't work.

So I'm kind of stuck on this one, if your own investigation leads to some concrete lead, I'd be happy to help.

PR #2020, opened 2019-01-04
Scrollspy-style updating of hash and TOC highlighting

edemaine:

## Summary

This PR adds two features:
1. Automatic updating of browser location's hash fragment according to what's on display (while the user scrolls, but without interfering with fragment while smooth scrolling / when clicking on a link).
2. Highlighting of the corresponding TOC items and its ancestors (similar to [Bootstrap's scrollspy](https://getbootstrap.com/docs/4.0/components/scrollspy/)).

To avoid slowdown while scrolling, I added a [simple throttling library](http://benalman.com/projects/jquery-throttle-debounce-plugin/), and replaced existing code where you were throttling inefficiently with `setInterval`.  This should improve idle performance while resizing identically.  (I wasn't totally sure whether to use `throttle` or `debounce`, but `throttle` seems like a fine choice.)

## Context

Follow-up to #2019.

For testing the second feature, it is useful to have a test with sticky TOC, but I didn't add one.  Let me know if you'd like a change or copy of `layout-table-of-contents-post`.


_sass/minimal-mistakes/_navigation.scss
```diff
@@ -507,6 +507,11 @@
     border-top-left-radius: $border-radius;
     border-top-right-radius: $border-radius;
   }
+
+  // Scrollspy marks toc items as .active when they are in focus
+  .active {
+    background: $active-color;
```
mmistakes: Switch to:

```scss
.active {
  @include yiq-contrasted($active-color);
}
```

_sass/minimal-mistakes/_variables.scss
```diff
@@ -61,6 +61,7 @@ $muted-text-color: mix(#fff, $text-color, 35%) !default;
 $border-color: $lighter-gray !default;
 $form-background-color: $lighter-gray !default;
 $footer-background-color: $lighter-gray !default;
+$active-color: mix($light-gray, $lighter-gray, 50%) !default;
```
mmistakes: Switch to the following and move after `$primary-color` line to avoid Sass error.

```scss
$active-color: mix(#fff, $primary-color, 80%) !default;
```

edemaine:
Fixed, thanks!  Sorry, I forgot about themes.

I guess another question is whether there should be an option to turn the auto-hash-changing functionality off.  But hopefully (most?) everyone wants it.

mmistakes:
I'm cool with leaving it on by default and not adding some sort of config. The theme is already a mess of configurations, don't need the added complexity :laughing: 

I could see someone potentially complaining about the size of `main.min.js` bulking up, but it looks like it's only ~1k or so to add these two often requested features.

sestus:
Hey, sorry to comment on this old issue, but is this working ? 

_Automatic updating of browser location's hash fragment according to what's on display (while the user scrolls, but without interfering with fragment while smooth scrolling / when clicking on a link)._

I don't see my browser's location being updated ( like https://coffeescript.org/ ).

edemaine:
It's been a while, but I believe Minimal Mistakes switched to using [Gumshoe](https://github.com/cferdinandi/gumshoe) instead of my custom code, and Gumshoe sadly doesn't support automatic hash fragment. If @mmistakes is supportive, I could try to add some code back that does just the hash fragment part, perhaps enabled by an option?

sestus:
@edemaine thanks for replying! That 'd be great. Do you have this already on your website? Could you point me to that code so that I could add this, in case @mmistakes doesn't want to add this back ?

iBug:
I haven't tested yet, but something like this should work:

```javascript
document.addEventListener("gumshoeActivate", function (event) {
  const target = event.target;
  const targetLink = target.querySelector(":scope > a[href]");
  if (!targetLink)
    return;
  const hash = targetLink.getAttribute("href");
  history.replaceState({}, "", hash);
});
```

[Gumshoe's custom events](https://github.com/cferdinandi/gumshoe#custom-events) are particularly useful here.

sestus:
@iBug legend! Just gave it a try and works like a charm. Many thanks!

PR #2021, opened 2019-01-04
Allow custom canonical url on a page-by-page basis

SeanKilleen:

## Summary

Currently, `seo.html` automatically calculates the canonical url.

This makes sense as a default 99% of the time, but occasionally may need to be overridden.

Example: I publish blog articles on my employer's blog occasionally cross-link them on my blog. In this case, to avoid an SEO hit to both my employer's and my SEO, I'd like to note that their URL is the canonical one. 

So, I'd like the ability to define the below in a post's YAML:

```yaml
canonical_url: https://my-employer.com/blog/blog-post-url.html
```

This PR updates `seo.html` to assign `canonical_url` to `page.canonical_url` if it exists.

## Testing
Pulled the repo and tested locally with a post file.

Before:

> image

After: 

> image


mmistakes:
I'm planning on doing away with the custom SEO include in the theme and using jekyll-seo-tag (see https://github.com/mmistakes/minimal-mistakes/issues/324). 

So this might be pointless in the interim.

SeanKilleen:
I see. Is that work close to completion? It seemed to have been ongoing for 2 years so I wasn't sure; figured this offered a quick solution in the interim.

mmistakes:
If it matches how jekyll-seo-tag expects a canonical URL to be defined I see no harm.

What I want is to have it at one way and that to be incompatible with jekyll-seo-tag. I've never used the feature with that plugin so not sure if how you're doing it is the save or not.

SeanKilleen:
👍 Makes perfect sense. I should have specified that I'm using the same syntax as `jekyll-seo-tag`. 

Per [that site's docs](https://jekyll.github.io/jekyll-seo-tag/advanced-usage#setting-customized-canonical-url):

> You can set custom Canonical URL for a page by specifying canonical_url option in page front-matter

So I think this change should use the same variable and be wholly compatible when it comes time to merge in that functionality.

PR #2022, opened 2019-01-07
Add support to customize `issue-term` for utterances comment provider

huan:
Current we hardcoded the `pathname` for `issue-term` when we are using utterances comment provider.

It would be better for let users customize it by setting a `site.comments.utterances.issue_term` in their `_config.yml`.


mmistakes:
Yes. If you use `remote_theme: "mmistakes/minimal-mistakes"` it'll pull whatever is on `master`.

huan:
That's great, thank you for the quick confirmation, and your theme is really awesome!

PR #2026, opened 2019-01-14
Add logo and title customization to the masthead

00-Evan:

## Summary

Adds an option for a logo image on the left side of the masthead. Also allows customization of the masthead title in case someone desires part or all of their title to be the logo.

A couple of things to note:
- The image is bounded by a max-height of 2rem, but can be smaller if the user gives a low resolution image. This would be useful for people who don't want their 32x32 logo to be scaled, or similar.
- Width of the image is not bounded, which would be helpful for people who want their logo to also contain their site name.
- I used margin-left 0 and margin-right 0.5rem as I think that's what looks best. If consistency with other masthead elements is more important though, it should probably be 0/1rem or 1rem/1rem
- The image shrinks down if it has to compete with the masthead title for space, which isn't great, but only happens with long titles or very wide images.

I fully admit that I'm not the most experienced with CSS, so if there are any ways to improve on this please let me know and I'll happily make adjustments.

Some example screenshots, using the test site:
[On Desktop](https://user-images.githubusercontent.com/2347603/51093128-0d63da80-176e-11e9-8596-b505262c1593.png)
[On Mobile](https://user-images.githubusercontent.com/2347603/51093129-0f2d9e00-176e-11e9-95f2-fa059f84648a.png)


mmistakes:
Looks great @00-Evan. I just added some documentation and a test example to `/test`.

justinrummel:
It appears that logo_path is always generated, thus htmlproofer is generating a BUNCH of errors because the source is blank.

```

**Comment by mmistakes** [comment]
Good catch @justinrummel 

The [Liquid logic](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/masthead.html#L11) probably needs to be changed to

```
{% if site.logo %}<a class="site-logo" href="{{ '/' | relative_url }}"><img src="{{ logo_path }}"></a>{% endif %}
```

**Comment by 00-Evan** [comment]
Whoops, thanks for the catch @justinrummel

I'll be at my computer in about 2 hours. Will get this fixed then.

**Comment by 00-Evan** [comment]
Okay @mmistakes , thanks!

**Comment by ashmaroli** [comment]
> The Liquid logic probably needs to be changed to...

@mmistakes Testing `site.logo` can lead to a false-positive when `site.logo` has been explicitly set to `''`
especially when `site.baseurl` is `nil` or `''`.

So IMO, the better route would be to test if `logo_path` is neither `nil` nor an empty string.

```liquid
{% if logo_path %}
  {% unless logo_path == "" %}
    ...
   {% endunless %}
{% endif %}
```
Alternatively, by using the `empty` Liquid builtin (since `{% capture %}` always returns a string):

```liquid
{% unless logo_path == empty %}
  ...
{% endunless %}
```

mmistakes:
Thanks for the suggestions @ashmaroli!

PR #2035, opened 2019-01-19
Add alternate text (alt attribute) to img tag with logo

torrocus:
This is a bug fix.

## Summary

It is good practice that all images have alternate text.
Image with the logo did not have that.
This is important for web accessibility and search engines.

## Context

No related with any GitHub issue.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -25,7 +25,6 @@ baseurl                  : # the subpath of your site, e.g. "/blog"
 repository               : # GitHub username/repo-name e.g. "mmistakes/minimal-mistakes"
 teaser                   : # path of fallback teaser image, e.g. "/assets/images/500x300.png"
 logo                     : # path of logo image to display in the masthead, e.g. "/assets/images/88x88.png"
-logo_alternate_text      : # alternate text which is displayed if logo image not exists, e.g. "Logo"
 masthead_title           : # overrides the website title displayed in the masthead, use " " for no title
 # breadcrumbs            : false # true, false (default)
 words_per_minute         : 200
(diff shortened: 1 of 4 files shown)


mmistakes:
This might be a bit redundant. Not **all** images need alt text, especially when the image is pure decoration. This is the case with the logo. 

There is already text describing the link, duplicating that with the logo may end up sounding strange from a screen reader when it repeats itself.

torrocus:
OK, you're partly right. However, if the image has no alternative text, the reader considers the path to the file as such and reads it. This is, in my opinion, a worse solution.

I noticed the problem using Google Chrome (v.71) -> Inspect (Ctrl + Shift + I) -> Audits -> Accessibility section. I will quote a fragment:
```
Image elements do not have [alt] attributes
Informative elements should aim for short, descriptive alternate text. Decorative elements can be ignored with an empty alt attribute. Learn more.
```
Here are links to more detailed descriptions: 
- https://dequeuniversity.com/rules/axe/2.2/image-alt
- https://www.w3.org/TR/WCAG20-TECHS/H67.html
- https://www.w3.org/WAI/tutorials/images/decorative/

So I changed the code. There will always be an alt attribute. If you do not provide alternative text, the value is empty.

mmistakes:
You're correct on the empty `alt` attribute. I had thought it was outputting `alt=""`, but looking at the PR I merged in they forgot the attribute entirely.

Not sure on the custom text conditional, but I'm OK with adding in just `alt=""` to silence an image path being read instead.

torrocus:
Can anyone decide itself? Is it good idea? People have different blogs, different needs. What do you think?

Of course, I can only add an empty alt attribute. I just want to know your opinion.

mmistakes:
My opinion is not to include it. The custom text is already present regardless if you use a logo or not.

In those cases where you use both you'd have it duplicated for assistive technologies.

Only use case I could see is if you have a blank title text and just show the logo.

I prefer simple over allowing configuring every possible thing. A user can always override an include to tailor it to fit their needs. I'm trying to balance this as their are already too many configs in theme and I've done a bad job of reigning them in.

PR #2053, opened 2019-02-05
update Font Awesome

coliff:
REF: https://fontawesome.com/changelog/latest

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -9,7 +9,7 @@
   {% endfor %}
 {% else %}
   <script src="{{ '/assets/js/main.min.js' | relative_url }}"></script>
-  <script defer src="https://use.fontawesome.com/releases/v5.6.0/js/all.js" integrity="sha384-z9ZOvGHHo21RqN5De4rfJMoAxYpaVoiYhuJXPyVmSs8yn20IE3PmBM534CffwSJI" crossorigin="anonymous"></script>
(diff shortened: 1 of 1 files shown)


_includes/scripts.html
```diff
@@ -9,7 +9,7 @@
   {% endfor %}
 {% else %}
   <script src="{{ '/assets/js/main.min.js' | relative_url }}"></script>
-  <script defer src="https://use.fontawesome.com/releases/v5.6.0/js/all.js" integrity="sha384-z9ZOvGHHo21RqN5De4rfJMoAxYpaVoiYhuJXPyVmSs8yn20IE3PmBM534CffwSJI" crossorigin="anonymous"></script>
+  <script defer src="https://use.fontawesome.com/releases/v5.7.1/js/all.js" integrity="sha384-eVEQC9zshBn0rFj4+TU78eNA19HMNigMviK/PU/FFjLXqa/GKPgX58rvt5Z8PLs7" crossorigin="anonymous"></script>
```
astrob0t:
```suggestion
  <script data-search-pseudo-elements defer src="https://use.fontawesome.com/releases/v5.7.1/js/all.js" integrity="sha384-eVEQC9zshBn0rFj4+TU78eNA19HMNigMviK/PU/FFjLXqa/GKPgX58rvt5Z8PLs7" crossorigin="anonymous"></script>
```

can we add the [`data-search-pseudo-elements`]() parameter to the script tag so as to enable CSS pseudo elements as we are using the JS framework.

astrob0t: @coliff CSS pseudo elements aren't that complicated. They are being used extensively in the theme, but using the font awesome icons as pseudo elements is not possible because it is disabled by default in the js framework. Since there is absolutely no overhead by enabling it but provides some additional functionalities, I was in favour of adding it. I will leave the decision upto @mmistakes.

mmistakes: If there is no overhead I'm OK with `data-search-pseudo-elements` being added.

PR #2055, opened 2019-02-07
exit search screen with Esc key

ermissa:

## Summary

This change allows you to close search screen with the ESC key. I was used to this shortcut and I thought it could be useful for others too.

## Context

It is not related to any GitHub issue.


mmistakes:
Think this is a good addition. Could you make the change in [`assets/js/_main.js`](https://github.com/mmistakes/minimal-mistakes/blob/master/assets/js/_main.js) as well? Having it in just the minified script will leave it open to being overwritten in the future when it's processed with `npm build:js`.

ermissa:
Oh sorry you are right. I changed _main.js. I also learned something new from my dummy mistake:D Thanks.

mmistakes:
Can you take care of the merge conflict on `assets/js/main.min.js`?
You'll need to run `npm run build:js` to compile the script, then commit the compiled version of `main.min.js` to the PR.

See the docs if you need help with the [build process](https://mmistakes.github.io/minimal-mistakes/docs/javascript/#build-process).

Also don't bump the version number. I'll take care of that in a future release.

PR #2121, opened 2019-04-07
POSH HTML5 <aside header titles >

inetbiz:

## Summary

Header titles do not start out at h4 for an outline inside certain sectioning tags such as `<aside>`.

## Context
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines

### DIFF
--- a/_includes/toc
+++ b/_includes/toc
@@ -1,6 +1,6 @@
 <aside class="sidebar__right">
 <nav class="toc" markdown="1">
-<header><h4 class="nav__title"><i class="fas fa-{{ include.icon | default: 'file-alt' }}"></i> {{ include.title | default: site.data.ui-text[site.locale].toc_label }}</h4></header>
+<header><h1 class="nav__title"><i class="fas fa-{{ include.icon | default: 'file-alt' }}"></i> {{ include.title | default: site.data.ui-text[site.locale].toc_label }}</h1</header>
 *  Auto generated table of contents
 {:toc .toc__menu}
 </nav>


mmistakes:
For SEO reasons this probably isn't a good idea as there is already a `h1` on the page and search engines generally advise you don't have more than one heading 1 on a page.

Also in the link above Mozilla even references that browsers and assistive technologies haven't fully implemented this HTML5 feature and following the document model is probably a good idea.

> **Important:** There are no implementations of the proposed outline algorithm in web browsers nor assistive technology; it was never part of a final W3C specification. Therefore the outline algorithm should not be used to convey document structure to users. Authors are advised to use heading rank (h1-h6) to convey document structure.

inetbiz:
Sections, Aside, Header are sub documents and still within a document, fall, automatically as subs. Using an h4 where you did just for stylization isn't semantically POSH. That much has been established. Run your docs through https://gsnedders.html5.org/outliner/ I suppose?

torrocus:
@inetbiz You have an incorrect closing tag: `</h1</header>`
In addition, I recommend reading the specification of HTML: https://www.w3.org/TR/WCAG20-TECHS/H42.html 
I suggest @mmistakes does not make this change.

inetbiz:
@torrocus I see your point. I can go back and refactor based on https://www.w3.org/TR/html5/sections.html#rank and refactor the PR? Discuss? And, realign header tags inside sectioning tags in html5

PR #2124, opened 2019-04-12
Fix Gumshoe related JavaScript error on pages without a TOC

Jeckari:
This is a bug fix.

## Summary

I've wrapped Gumshoe in an if statement. This prevents a console error on pages without a TOC. 

The console error occurred when you scroll to the bottom of a page without a TOC.


mmistakes:
Thanks for discovering and fixing this bug @Jeckari.

Issue #2173, opened 2019-05-30
Display site.description in masthead


jdeluyck:
## Summary

Allow printing of the subtitle / description of the site below the site title

## Motivation

It's just a nice featue that I believe more sites can use. It's like a subnote under your main title, giving you the ability to add an 'afterthought'.

## Drawbacks

Can't really think of one.

## Code:

I added to main.css:
````
$type-size-9: 0.350em !default; // ~4px

/* Site Subtitle */
.site-subtitle {
  font-size: $type-size-9;
}
````

to masthead
````
        <a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}
        <div class="site-subtitle">{{ site.description }}</div></a>
````

and I guess this could be made optional - for my own use I've just included it as is.

iBug:
Good idea. But what about submitting a PR for this?

jdeluyck:
Sure :) I wasn't sure if you'd be interested.

mmistakes:
@jdeluyck If it's quick to do a PR I'll take a look, but no promises I'll merge it in. I'm trying to reduce the amount optional stuff since the theme is getting really bloated with these sorts of things.

They're easy enough to do custom `_includes` that override the bundled theme's if you wish to customize.

PR #2205, opened 2019-07-03
Display site subtitle in masthead

jdeluyck:
## Summary

Displays `site.subtitle` in masthead under the site title.

## Context

Related to issue #2173

### DIFF
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -11,7 +11,12 @@
         {% unless logo_path == empty %}
           <a class="site-logo" href="{{ '/' | relative_url }}"><img src="{{ logo_path | relative_url }}" alt=""></a>
         {% endunless %}
-        <a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}</a>
+        <a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}
+	{% if site.subtitle == empty %}
+	  </a>
+	{% else %}
+	  <div class="site-subtitle">{{ site.description }}</div></a>
+        {% endif %}
         <ul class="visible-links">
           {%- for link in site.data.navigation.main -%}
             {%- if link.url contains '://' -%}
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -12,11 +12,7 @@
           <a class="site-logo" href="{{ '/' | relative_url }}"><img src="{{ logo_path | relative_url }}" alt=""></a>
         {% endunless %}
         <a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}
-        {% if site.subtitle == empty %}
-	  </a>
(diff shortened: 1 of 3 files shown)


_includes/masthead.html
```diff
@@ -11,7 +11,12 @@
         {% unless logo_path == empty %}
           <a class="site-logo" href="{{ '/' | relative_url }}"><img src="{{ logo_path | relative_url }}" alt=""></a>
         {% endunless %}
-        <a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}</a>
+        <a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}
+        {% if site.subtitle == empty %}
```
mmistakes: I think this can be written much simpler like this:

```html
<a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}
  {% if site.subtitle %}<div class="site-subtitle">{{ site.description }}</div>{% endif %}
</a>
```

_sass/minimal-mistakes/_masthead.scss
```diff
@@ -53,6 +53,10 @@
   z-index: 20;
 }
 
+.site-subtitle {
+  font-size: $type-size-9;
```
mmistakes: This font-size is way too small, you can't read it. It'll also ding you on accessibility. I'd stick with `$type-size-8` as the smallest size you'd want to go.

jdeluyck: Hm, ok. On my own site it actually looks ok, as I really just want it as a subtext.
But sure.

_includes/masthead.html
jdeluyck: Duh. Ofcourse.

_includes/masthead.html
```diff
@@ -11,7 +11,9 @@
         {% unless logo_path == empty %}
           <a class="site-logo" href="{{ '/' | relative_url }}"><img src="{{ logo_path | relative_url }}" alt=""></a>
         {% endunless %}
-        <a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}</a>
+        <a class="site-title" href="{{ '/' | relative_url }}">{{ site.masthead_title | default: site.title }}
+        {% if site.subtitle %}<div class="site-subtitle">{{ site.description }}</div>{% endif %}
```
mmistakes: Any particularly reason you're using `site.subtitle` as the conditional, but displaying `site.description`? Wouldn't it make more sense to output `site.subtitle`?

A site description and site subtitle might be different. I would favor keeping things consistent.
If your'e checking for `site.subtitle` output that if `true`. Otherwise use `{% if site.description%}<div class="site-description">{{ site.description }}</div>{% endif %}` makes more sense to me.

Though that might not be intended by people who have a description filled out for SEO `<meta>` tag reasons but don't want it shown in masthead.

PR #2211, opened 2019-07-14
Fixes aria issues on search form

JimDrury:
This is a bug fix.

## Summary
Allows search input to have aria-label defined through language conflict, which resolves the A11y issue of an input without a label.

Also removes aria-placeholder from the search label, as per [ARIA](https://www.w3.org/TR/html-aria/) 

```
Use the placeholder attribute on any element that is allowed the placeholder attribute in HTML5
Only use the aria-placeholder attribute on elements that are not allowed to have a placeholder attribute in HTML5.
```

## Context
Resolves issue: [2180](https://github.com/mmistakes/minimal-mistakes/issues/2180)

### DIFF
--- a/_includes/search/search_form.html
+++ b/_includes/search/search_form.html
@@ -11,7 +11,7 @@
     <div id="results" class="results"></div>
   {%- when "google" -%}
     <form onsubmit="return googleCustomSearchExecute();" id="cse-search-box-form-id">
-      <label  class="sr-only" for="cse-search-input-box-id">
+      <label class="sr-only" for="cse-search-input-box-id">
         {{ site.data.ui-text[site.locale].search_label_text | default: 'Enter your search term...' }}
       </label>
       <input type="search" id="cse-search-input-box-id" class="search-input" tabindex="-1" placeholder="{{ site.data.ui-text[site.locale].search_placeholder_text | default: 'Enter your search term...' }}" />
--- a/_includes/search/search_form.html
+++ b/_includes/search/search_form.html
@@ -2,11 +2,19 @@
 {%- assign search_provider = site.search_provider | default: "lunr" -%}
 {%- case search_provider -%}
   {%- when "lunr" -%}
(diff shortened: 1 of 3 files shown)


mmistakes:
Instead of using `aria-label` why not use an actual `<label>` element in combination with the `input`?

https://github.com/mmistakes/minimal-mistakes/pull/2181#issuecomment-500124013

I suggested that in the previous PR to fix the A11y issue.

JimDrury:
Fair point tbh
I’ll take a look shortly

_includes/search/search_form.html
```diff
@@ -2,11 +2,19 @@
 {%- assign search_provider = site.search_provider | default: "lunr" -%}
 {%- case search_provider -%}
   {%- when "lunr" -%}
-    <input type="search" id="search" aria-label="{{ site.data.ui-text[site.locale].search_label_text | default: 'Enter your search term...' }}" class="search-input" tabindex="-1" placeholder="{{ site.data.ui-text[site.locale].search_placeholder_text | default: 'Enter your search term...' }}" />
+    <form>
```
mmistakes: If you wrap the input with a `<form></form>` element, hitting <kbd>Enter</kbd> tries to submit the form, which reloads the page. This is why it's not included currently since it's not actually submitting anything.

Though I'm not sure if that's correct semantically or not. Can you have an `input` and `label` without inclosing them in a `form`?

If you do, then we might need to add some JavaScript to disable the default behavior that reloads the page when hitting <kbd>Enter</kbd>.

mmistakes: Looks like it's cool to have them outside of a `form` element.

https://stackoverflow.com/questions/8568660/is-there-any-danger-to-using-input-fields-outside-without-forms-in-html-javascri

JimDrury: Resolved this now, it's inside a form, i've gone for a slight ghetto onkeydown solution to prevent keypress, didnt seem to warrant an external file.

mmistakes:
Thanks for making the suggested changes, everything looks good to me.

PR #2217, opened 2019-07-22
Added image size control

mmorton234:

## Summary

This will add the ability to control the max size of images in a feature row element
This is done by adding something along the lines of (below) to your feature row element declaration
image_max_size: 100% | image_max_size: 10% | image_max_size: 2em | image_max_size: 150px

## Drawback
if someone wishes to add an image that has is small in size and want it to take up the whole space they will have to specify that the max size should be 'x' number of pixels with x being the max size they want 

## Context
This was made to fix an issue of images appearing too large and out of place on small screens and very large screens when the feature rows are on left or right 

## Improvements 
This was something I was unable to do but would improve the functionality of the feature
(sorry new to html and css tried but couldn't get anything working)
If the addition of vertical and horizontal centering of the image with a changed size was added it would make smaller images appear better suited when used in left and right feature rows


mmorton234:
needed to work off of that fork will summit another one sorry

PR #2230, opened 2019-08-09
Fix default site.author in seo.html

zirius:
This is a bug fix.

## Summary
SEO now shows author tag properly


mmistakes:
Great catch! Thanks for the PR.

PR #2244, opened 2019-08-29
Fix http links to use https

d-Rickyy-b:
Every site should support https by now. Better to link the https sites right away. Both sites do support https at the time I checked.

## Summary
I changed the protocol of some links given in the comments of the config file from http to https since all sites should make use of https in 2019

## Context
This PR is not related to any GitHub issues.


mmistakes:
Looks good. Thanks!

PR #2274, opened 2019-10-04
delete unused files

junsu-ku:


junsu-ku:
oops sorry i made a mistake this is not my page...

PR #2279, opened 2019-10-13
Add automated page and post makers

raviriley:

## Summary

I built scripts to automatically create pre-formatted post and page markdown files. 
I'm not sure if this is the right place to add them, but I think these could be very useful to people.

## Context

Not related to any issues.

### DIFF
--- a/docs/_pages/newpage.sh
+++ b/docs/_pages/newpage.sh
@@ -1,0 +1,12 @@
+echo "page title: (ex: My Projects)"
+read title
+#echo "filename is lowercase and -'d title, i.e. my-portfolio"
+name=${title,,} #to lower case
+file=${name// /-}.md #replace spaces with - and adds .md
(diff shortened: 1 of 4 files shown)


mmistakes:
Think this is better left out of the theme. She’ll scripts don’t have universal support in all environments... especially on Windows.

There are also many plugins, Gems, projects, and other scripts that do similar things in Jekyll.

raviriley:
I understand if you don't want to include it. 

I am on Windows, and made this so it would run on Windows or a UNIX OS. It worked on my friend's Mac and a Ubuntu server. Most Windows users with GitHub have Git and Git Bash installed, which means they can do bash scripting. 

Maybe a link to my scripts could be added in to documentation somewhere?

mmistakes:
Would probably do more good linking to them on Jekyll docs resource page than the theme’s.

https://jekyllrb.com/resources/

PR #2292, opened 2019-10-20
Added more advanced example

lsolesen:

## Summary

I have added a form example using some of the stylesheets.

The example reveals a couple of things.

- Legends for fieldsets containing radio and checkboxes should probably be left aligned?
- Fieldsets are not separated in the layout
- It is not clear what form-group is used for
- It would be cool if it was possible to make a little more structured layout when using input-mini and input-small

Right now it looks like this on /markup/markup-html-tags-and-formatting/:

Screenshot

If it was possible to use the classes for some sort of alignment - e.g.

Screenshot (22)

I think I more or less captured all the classes found in the code, but might be missing some.

### DIFF
--- a/_sass/minimal-mistakes/_forms.scss
+++ b/_sass/minimal-mistakes/_forms.scss
@@ -277,8 +277,7 @@ select:focus {
 }
 
 .form-inline .radio,
-.form-inline .checkbox,
-.form-inline .radio {
+.form-inline .checkbox {
   padding-left: 0;
   margin-bottom: 0;
   vertical-align: middle;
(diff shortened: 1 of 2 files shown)


lsolesen:
@mmistakes I added a little more markup trying to guess what the preferred markup for forms should be. I think it would be awesome to add the preferred way of marking up forms into the docs.

iBug:
LGTM, why stale?

lsolesen:
@mmistakes I am not quite sure why this ended up as stale. I think it is a beneficial addition to the docs?

Issue #2261, opened 2019-09-12
Suggestion for docs: Link posts to their markdown in Github


brigitte-youitv:
## Motivation

I love all the sample posts you have. They're amazing. And it's great to see that they use all kinds of different helpers and examples. 

What I find myself doing again and again though, is having to open the github project, locate each sample file, and see how you did something.

For example, yesterday I wanted to know what class or helper or *thing* you used to align the images here: https://mmistakes.github.io/minimal-mistakes/markup/markup-image-alignment/ Although it's not hard to go find it in github, it's tedious.

## Suggestion

Consider adding a "see source" link to the top or bottom of pages that takes the reader to the source markdown file in github (raw view, or file view) so that it's quick and easy to see how you accomplished something.

PR #2299, opened 2019-10-28
Issue #2261 - Suggestion for docs: Link posts to their markdown in GitHub

sprak3000:

## Summary

Add links to the post examples that go to the raw content of the post on GitHub. Allows one to see the actual markdown, etc. used on the page.

## Context

This is related to issue #2261. I have also found myself going to the repo to see the actual source of the post to determine what markdown is being used, etc.

### DIFF
--- a/_layouts/single.html
+++ b/_layouts/single.html
@@ -42,6 +42,9 @@ layout: default
             </nav>
           </aside>
         {% endif %}
+        {% if site.show_edit_on_github_link %}
+        <p><a href="{% github_edit_link %}">{{ site.edit_on_github_link_text }}</a></p>
+        {% endif %}
         {{ content }}
         {% if page.link %}<div><a href="{{ page.link }}" class="btn btn--primary">{{ site.data.ui-text[site.locale].ext_link_label | default: "Direct Link" }}</a></div>{% endif %}
       </section>
--- a/_config.yml
+++ b/_config.yml
@@ -28,6 +28,8 @@ teaser                   : # path of fallback teaser image, e.g. "/assets/images
 logo                     : # path of logo image to display in the masthead, e.g. "/assets/images/88x88.png"
 masthead_title           : # overrides the website title displayed in the masthead, use " " for no title
 # breadcrumbs            : false # true, false (default)
+show_edit_on_github_link : false # true, false (default)
+edit_on_github_link_text : "Edit this page on GitHub"
 words_per_minute         : 200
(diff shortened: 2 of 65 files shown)


sprak3000:
I feel this would be beneficial given you often need to look up the front matter or markup to understand how to format your own pages in the style of the examples. Having things linked for easy reference eases the barrier for learning how to use the theme effectively.

mmistakes:
It's a good suggestion but it would be a maintenance hell manually adding the link to all documents. A better way would be to leverage GitHub pages Jekyll filter for doing this.

https://ben.balter.com/2015/09/13/github-pages-edit-button/

Ideally it would be a feature added to the theme as well that could be toggled on so other sites built on MM could utilize it... not just the theme's documentation.

sprak3000:
Took your suggestion and ran with it for at least a first pass.

_layouts/single.html
```diff
@@ -42,6 +42,9 @@
             </nav>
           </aside>
         {% endif %}
+        {% if site.show_edit_on_github_link %}
+        <p><a href="{% github_edit_link %}">{{ site.edit_on_github_link_text }}</a></p>
```
lsolesen: I think you should consider building the link like this instead: 

    https://github.com/{{site.repository}}/edit/gh-pages/{{ page.path }}

Then github is not needed as a dependency and it will still work.

_layouts/single.html
```diff
@@ -42,6 +42,9 @@
             </nav>
           </aside>
         {% endif %}
+        {% if site.show_edit_on_github_link %}
```
lsolesen: I think it should also be possible to override this per page, as you might have collections of single pages, that you want collaborators on, but others you do not.

_layouts/single.html
mmistakes: @Isolesen - don't think `https://github.com/{{site.repository}}/edit/gh-pages/{{ page.path }}` will work as what happens when you don't use the `gh-pages` brand? The path will vary.

By not needing GitHub as a dependency... are you referring to GitHub Pages? Not sure if that's a big deal. I suppose `{% github_edit_link %}` could be replaced with a `site.` variable that you could define the base path. But that adds another layer of complexity and something else for the user to customize.

I really don't want to overthink this as I wasn't completely sold on the feature needing to be in the core theme. Only really considered it since `{% github_edit_link %}` made things easier to generate the link.

PR #2328, opened 2019-11-21
Fix unreadable code blocks in notices

iBug:
This is a bug fix.

Before this patch, code blocks `<pre><code>` in notices are hardly readable:

> before

After this patch, they look certainly better!

> after


mmistakes:
Looks good, thanks!

PR #2334, opened 2019-11-27
If URL is external links to open link in new tab

llovvoll:

## Summary

I think readers always leave the current page if they click on the URL. It is a very bad experience, so I added the judgment (hostname), if the URL is an external link, open link in new tab, and if it is an internal link, change the current page

P.S. English is not my primary language, please forgive me

## Context

Not related to any existing GitHub issue.

## Demo
[https://ovvo.cc/my-develop-settings-when-reinstall/](https://ovvo.cc/my-develop-settings-when-reinstall/)


mmistakes:
Forcing external links into a new window is a bad user experience in my opinion for all the reasons mentioned in this article.

PR #2339, opened 2019-12-04
Fix schema.org dates to ISO-8601

perelo:
This is a bug fix.

## Summary
It seems that [schema.org's *date* data type](https://schema.org/Date) should be ISO-8601. So I updated `datePublished` and `dateModified` format to [`date_to_xmlschema`](https://jekyllrb.com/docs/liquid/filters/).

### DIFF
--- a/_includes/comment.html
+++ b/_includes/comment.html
@@ -13,7 +13,7 @@
     <p class="comment__date">
       {% if include.date %}
         {% if include.index %}<a href="#comment{{ include.index }}" itemprop="url">{% endif %}
-        <time datetime="{{ include.date | date_to_xmlschema }}" itemprop="datePublished">{{ include.date | date: "%B %d, %Y at %I:%M %p" }}</time>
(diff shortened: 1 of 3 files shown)


_includes/comment.html
```diff
@@ -13,7 +13,7 @@ <h3 class="comment__author" itemprop="author" itemscope itemtype="https://schema
     <p class="comment__date">
       {% if include.date %}
         {% if include.index %}<a href="#comment{{ include.index }}" itemprop="url">{% endif %}
-        <time datetime="{{ include.date | date_to_xmlschema }}" itemprop="datePublished">{{ include.date | date: "%B %d, %Y at %I:%M %p" }}</time>
+        <time datetime="{{ include.date | date_to_xmlschema }}" itemprop="datePublished">{{ include.date | date_to_xmlschema }}</time>
```
mmistakes: `<time>` is already using `date_to_xmlschema` for the `datetime` attribute. My understanding the time between `<time></time>` is the human readable format and not used.

perelo: You're [right](https://schema.org/docs/gs.html#advanced_dates) ! I reverted it.

PR #2340, opened 2019-12-04
Specify margin-top for archive_item-title

statzhero:
## Summary 
Don't know if this is an idiosyncratic edit for my local version of MM only (made some design changes), but it fixed my issue of too much top margin for the "recent post" section. Please ignore if not relevant (but do explain as I'm curious to learn). Thanks!

This is a bug fix.

## Context
No

### DIFF
--- a/_sass/minimal-mistakes/_archive.scss
+++ b/_sass/minimal-mistakes/_archive.scss
@@ -44,6 +44,7 @@
 }
 
 .archive__item-title {
+  margin-top: 0.5em;
   margin-bottom: 0.25em;
   font-family: $sans-serif-narrow;
   line-height: initial;


mmistakes:
Where are you using it as depending on the context `margin-top: 0` is applied.

For example on this page:
- https://mmistakes.github.io/minimal-mistakes/year-archive/

statzhero:
This is for, I believe, the `h3` posts in archive-single.html

I should explain that I'm not really sure what's going on, just that it worked.

mmistakes:
Do you have an example? Because the link above is using archive single and it doesn't have the padding issue.

Unless it's just a matter of preference where you don't like the spacing. If that's the case then this is something better left in your personal fork of the theme.

statzhero:
You're most probably right, so just ignore it. I added the below to a `layout: categories` page, but can't remember how exactly I got there. 

```
<h3 class="archive__subtitle">{{ site.data.ui-text[site.locale].recent_posts | default: "Recent Posts" }}</h3>

{% for post in site.posts limit: 3 %}
  {% include archive-single.html %}
{% endfor %}
```

mmistakes:
Not sure what you're goal was, but `layout: categories` is meant to show posts from a specific category, not recent posts. `layout: home` is better for that.

And the reason your margin's seemed large between the titles is you need to wrap it in a section tag with a specific class (`.taxonomy__section`). This applies `margin-top: 0` to the titles.

See below:

```html
<section id="{{ category[0] | slugify | downcase }}" class="taxonomy__section">
  <h2 class="archive__subtitle">{{ category[0] }}</h2>
    <div class="entries-{{ page.entries_layout | default: 'list' }}">
      {% for post in category.last %}
        {% include archive-single.html type=page.entries_layout %}
      {% endfor %}
    </div>
    <a href="#page-title" class="back-to-top">{{ site.data.ui-text[site.locale].back_to_top | default: 'Back to Top' }} &uarr;</a>
</section>
```

https://github.com/mmistakes/minimal-mistakes/blob/master/_layouts/categories.html#L31-L39

statzhero:
thank you!

PR #2342, opened 2019-12-09
Use HTML meta keywords if available from page.keywords

iBug:

Generate a `<meta name=keywords>` tag if keywords for a page is available. Good for SEO.

### DIFF
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -25,6 +25,10 @@
   {%- assign seo_description = seo_description | markdownify | strip_html | strip_newlines | escape_once -%}
 {%- endif -%}
 
+{%- if page.keywords -%}
+  {%- assign seo_keywords = page.keywords | join: ',' -%}
+{%- endif -%}
+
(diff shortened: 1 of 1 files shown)


mmistakes:
I disagree with this being good for SEO. It's long since been regarded as not a good idea.

https://yoast.com/meta-keywords/

marklchaves:
At the risk of being too _facetious_, I just want to say, I love Yoast.

Since Yoast is for _mass consumption_, most people believe everything they say. This _fact_ helps me with my SEO, because I do the things (like _meta_ keywords--I've been using since '96) that Yoast doesn't recommend. 

BTW, one of my sites has been on Google SERP 1 position 1 for the three years it's been alive. And, without ever _touching_ Yoast. ;-)

Just my two cents.

Thanks!

mmistakes:
It's not just Yoast offering this advice. It's my understanding search engines ignore it because it's been so abused in the past and offers no value.

It's your actual page content that is used for search. Closing as this isn't something I'd like to add to the theme.

PR #2345, opened 2019-12-13
Hide posts with [hidden: true] from listings

iBug:

Hide posts which have `hidden: true` in their front matters from collection / category / tag listings, as is already done with `home` layout. This works for listing pages with `layout: collection OR category OR tag`.

### DIFF
--- a/_includes/documents-collection.html
+++ b/_includes/documents-collection.html
@@ -15,5 +15,7 @@
 {% endif %}
 
 {%- for post in entries -%}
-  {% include archive-single.html %}
+  {%- unless post.hidden -%}
+    {% include archive-single.html %}
+  {%- endunless -%}
 {%- endfor -%}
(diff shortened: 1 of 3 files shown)


torrocus:
Could you explain why exactly this functionality should be used?

Maybe you want to have posts you're still working on and don't want to publish yet?
This functionality is already in Jekyll. Use the _draft folder and the --draft option (https://jekyllrb.com/docs/posts/#drafts)


Also your way is fake-hiding. First of all, the post will still be public and can be accessed. Secondly, external gems do not know the "hidden" attribute and it will not work. RSS is a good example. There is still your "hidden post".

iBug:
@torrocus It's the default behavior for both `jekyll-paginate` and `jekyll-paginate-v2`, where posts with this attribute aren't listed in indices. Other than that this setting has no effect. But still, I'm relying on this to have pages that are "visible only if you know the link". Adding the functionality to other listings than the main index makes it more consistent.

iBug:
not stale

iBug:
not stale again

PR #2351, opened 2019-12-19
Fix staticman v2/v3 comments not showing

akhy:
This is a bug fix.

## Summary

According to the most recent docs, `branch` and `endpoint` should be inside `site.comments.staticman` config instead of `site.staticman`.

## Context

https://mmistakes.github.io/minimal-mistakes/docs/configuration/#staticman-v3

### DIFF
--- a/_includes/comments.html
+++ b/_includes/comments.html
@@ -12,7 +12,7 @@
       <section class="fb-comments" data-href="{{ page.url | absolute_url }}" data-mobile="true" data-num-posts="{{ site.comments.facebook.num_posts | default: 5 }}" data-width="100%" data-colorscheme="{{ site.comments.facebook.colorscheme | default: 'light' }}"></section>
     {% when "staticman_v2" %}
       <section id="static-comments">
-        {% if site.repository and site.staticman.branch %}
+        {% if site.repository and site.comments.staticman.branch %}
           <!-- Start static comments -->
           <div class="js-comments">
             {% if site.data.comments[page.slug] %}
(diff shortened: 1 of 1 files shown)


deadlydog:
FYI @mmistakes  this was actually a breaking change. I had my blog's `_config.yml` set to use the latest commit via:

```yaml
remote_theme: mmistakes/minimal-mistakes
```

And never noticed until today that comments were no longer showing up on my blog. Before finding this PR, a quick fix was to do:

```yaml
remote_theme: mmistakes/minimal-mistakes@26c1989
```

Now that I see the actual change you applied, I'm sure I can update my `_config.yml` appropriately to have it place the variables in their new expected place. The fix was simply to indent the lines:

```yaml
staticman:
  branch: "master"
  endpoint: "https://MyEndpoint.herokuapp.com/v3/entry/github/"
```

to (added 2 spaces in front of each line):

```yaml
  staticman:
    branch: "master"
    endpoint: "https://MyEndpoint.herokuapp.com/v3/entry/github/"
```

I'm posting the fix for others who come across this issue and find this PR.

Just thought I'd mention this, as you may not have realized you were making a breaking change here, and may want to communicate it somehow in the ReadMe for others that discover their comments have suddenly gone missing.

mmistakes:
That’s one of the big drawbacks of using remote themes. Every commit to the repo has the possibility of breaking your site. This is why I encourage installing it by locking at specific versions or commits... as you’ve pointed out above.

deadlydog:
Yeah, I'll be locking it via commit message going forward. I also found a bug introduced by this PR and created the PR above to address it :) Thanks for your great theme and hard work!

PR #2373, opened 2020-01-14
Adjust comments to be compatible with compress_html

gamue:
This is a bug fix. 

## Summary

Since enabling compress_html at my page the disqus comments haven't been shown. When looking at the compressed page code I've saw that the `//` comments are also disabling the code after that as compression removed the line-breaks.
Using `/* */` for comments clearly wraps the comment and doesn't deactivate any code afterwards.

## Context

- see also Restrictions at https://jch.penibelst.de/ 
- I added this changes to my repo, so you could see it running (and working at) https://gamue.de/

### DIFF
--- a/_includes/comments-providers/disqus.html
+++ b/_includes/comments-providers/disqus.html
@@ -1,10 +1,10 @@
 {% if site.comments.disqus.shortname %}
   <script>
     var disqus_config = function () {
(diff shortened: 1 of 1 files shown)


mmistakes:
Looks good, thanks!

PR #2378, opened 2020-01-21
Allow using theme without pagination

mohd-akram:

## Summary

It allows using the theme without pagination.

## Context

Right now if you get this theme, it will not show any posts on the homepage unless you add an `index.html` file and enable pagination in `_config.yml`. This change reduces this friction and some head-scratching as to why no posts are showing on the homepage.


mohd-akram:
Sure, updated.

PR #2332, opened 2019-11-26
Update masthead.html

dmitrypv:
fix image path in masthead for relative url

This is a bug fix. 

## Summary
Absolute/relative path check {% if site.logo contains "://" %} for logo image must prepend the baseurl in case of relative path.

## Context
If you specify 'logo: "/assets/images/logoexample.png' in _config.yml, logoexample.png will not be shown in masthead unless absolute path is used.

PR #2385, opened 2020-01-29
Regression for mmistakes#2332

iBug:
This is a bug fix.

## Summary

On the merged commit of #2332 (abf29432f6e4d75a13e94a20d8a8fc93953ee9e7), there's already

<https://github.com/mmistakes/minimal-mistakes/blob/abf29432f6e4d75a13e94a20d8a8fc93953ee9e7/_includes/masthead.html#L12>

Emphasis:

```html
<img src="{{ logo_path | relative_url }}" alt="">
                       ^^^^^^^^^^^^^^
````

so the extra `relative_url` is the culprit for the duplicated base path for masthead logo when `site.baseurl` isn't resolved to `/`.

This PR resolves that bug by removing the extra `relative_url`. On a side note, I wouldn't blame the original PR suggester because the initial `{% if site.logo contains "://" %}` is rather misleading per se.

## Context

#2332 Comment: <https://github.com/mmistakes/minimal-mistakes/pull/2332#issuecomment-579856138>

cc @Geekdude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Bug Fixes
+
+- Fix repeated site base path for masthead logo. [#2385](https://github.com/mmistakes/minimal-mistakes/pull/2385)
+
 ## [4.18.1](https://github.com/mmistakes/minimal-mistakes/releases/tag/4.18.1)
 
 ### Bug Fixes
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -1,8 +1,4 @@
-{% if site.logo contains "://" %}
-  {% capture logo_path %}{{ site.logo }}{% endcapture %}
-{% else %}
-  {% capture logo_path %}{{ site.logo | relative_url }}{% endcapture %}
-{% endif %}
+{% capture logo_path %}{{ site.logo }}{% endcapture %}
 
 <div class="masthead">
   <div class="masthead__inner-wrap">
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by mmistakes** [comment]
Can you verify this works with externally hosted images? That is what the `//:` check is for. There are users who don't host the images on their site (not sure why you'd want to do that but they do...).

So we need to be able to do `logo: http://cdn.whatever.com/my-external-logo.png` and `/assets/images/my-logo.png`

**Comment by iBug** [comment]
@mmistakes I've verified locally with both Jekyll 3.8.6 and Jekyll 4.0. Here are the results:

```text
baseurl: /test
{{ "abcd" | relative_url }} -> /test/abcd
{{ "https://google.com/abcd" | relative_url }} -> https://google.com/abcd
```

So your worry isn't practical, Jekyll has already taken care of it with the `relative_url` filter.

mmistakes:
OK cool. `relative_url` didn't use to work that way which is why that extra check was needed. Guess if they fixed that all the other spots in the theme where there's an absolute URL check could be removed. Think it's mostly in image includes... hero, galleries maybe, etc.

iBug:
@mmistakes Running `git blame` over the Jekyll code base shows <https://github.com/jekyll/jekyll/commit/a66c478> (merge of jekyll/jekyll#6490) on November 5, 2017. The milestone suggests that the change was carried out with the release of [Jekyll 3.7.0](https://github.com/jekyll/jekyll/releases/tag/v3.7.0). So if you want to utilize this new feature, you might want to update this badge in `README.md`:

<https://github.com/mmistakes/minimal-mistakes/blob/1799ddd6a53198a379244e807ab3be2bdf027353/README.md#L4>

PR #2387, opened 2020-01-29
Use relative_url and absolute_url where possible

iBug:

## Summary

Since Jekyll 3.7.0 the `relative_url` filter does nothing when given an absolute URL, which eliminates the need to manually check for `url contains "://"`. This saves a lot of effort when coding.

A similar behavior for the `absolute_url` filter had already been introduced earlier in jekyll/jekyll#5789 and was released with Jekyll 3.5.

I ran both `grep -nrF 'contains "://"'` and `grep -nrF "contains '://'"` (single- vs. double-quotes) and reviewed every single occurrence of them, and refactored wherever possible / makes sense.

Jekyll 3.6 would no longer be compatible with this theme after merging this PR.

## Context

https://github.com/mmistakes/minimal-mistakes/pull/2385#issuecomment-579882236

### DIFF
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -8,8 +8,7 @@
 			{% assign author_src = author.avatar | relative_url %}
 
       {% if author.home %}
-				{% assign author_link = author.home | relative_url %}
-        <a href="{{ author_link }}">
+        <a href="{{ author.home | relative_url }}">
           <img src="{{ author_src }}" alt="{{ author.name }}" itemprop="image">
         </a>
       {% else %}
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -5,14 +5,12 @@
 
   {% if author.avatar %}
     <div class="author__avatar">
-			{% assign author_src = author.avatar | relative_url %}
-
       {% if author.home %}
         <a href="{{ author.home | relative_url }}">
-          <img src="{{ author_src }}" alt="{{ author.name }}" itemprop="image">
+          <img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image">
         </a>
       {% else %}
-        <img src="{{ author_src }}" alt="{{ author.name }}" itemprop="image">
+        <img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image">
       {% endif %}
     </div>
   {% endif %}
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -5,18 +5,10 @@
 
   {% if author.avatar %}
     <div class="author__avatar">
-      {% if author.avatar contains "://" %}
-        {% assign author_src = author.avatar %}
-      {% else %}
-        {% assign author_src = author.avatar | relative_url %}
-      {% endif %}
+			{% assign author_src = author.avatar | relative_url %}
 
       {% if author.home %}
-        {% if author.home contains "://" %}
-          {% assign author_link = author.home %}
-        {% else %}
-          {% assign author_link = author.home | relative_url %}
-        {% endif %}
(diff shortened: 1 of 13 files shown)


_includes/author-profile.html
```diff
@@ -208,7 +200,7 @@ <h3 class="author__name" itemprop="name">{{ author.name }}</h3>
               <i class="fab fa-fw fa-youtube" aria-hidden="true"></i> YouTube
             </a>
           </li>
-        {% else author.youtube %}
```
iBug: This is a side fix that I did when I scanned through the file. Correct me if I'm wrong.

mmistakes: Might not need the `assign` anymore either. Looks like `author.home` is only being used the one time so adding `relative_url` directly to that should be fine.

```
{% if author.home %}
  <a href="{{ author.home | relative_url }}">
  ...
```

iBug:
Any ideas? @mmistakes

_includes/author-profile.html
```diff
@@ -5,19 +5,10 @@
 
   {% if author.avatar %}
     <div class="author__avatar">
-      {% if author.avatar contains "://" %}
-        {% assign author_src = author.avatar %}
-      {% else %}
-        {% assign author_src = author.avatar | relative_url %}
-      {% endif %}
+			{% assign author_src = author.avatar | relative_url %}
```
mmistakes: Can further simplify by removing this `assign` and add the `relative_url` filter directly to `author.avatar` in the image elements.

```html
<img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image">
```

_includes/author-profile.html
```diff
@@ -5,19 +5,10 @@
...
 
       {% if author.home %}
-        {% if author.home contains "://" %}
-          {% assign author_link = author.home %}
-        {% else %}
-          {% assign author_link = author.home | relative_url %}
-        {% endif %}
-        <a href="{{ author_link }}">
+        <a href="{{ author.home | relative_url }}">
           <img src="{{ author_src }}" alt="{{ author.name }}" itemprop="image">
```
mmistakes: Replace `{{ author_src }}` with `{{ author.avatar | relative_url }}`

_includes/nav_list
```diff
@@ -9,11 +9,7 @@
       <li>
         {% if nav.url %}
           {% comment %} internal/external URL check {% endcomment %}
```
mmistakes: Here's another one. Can remove the `{% comment %}` and `{% assign nav_url %}` and go with just `{{ nav-url | relative }}` instead of `{{ nav_url }}`.

iBug:
@mmistakes I assume you received notifications for the last commit?

_includes/page__hero.html
```diff
@@ -1,20 +1,6 @@
...
-{% else %}
-  {% capture cta_path %}{{ page.header.cta_url | relative_url }}{% endcapture %}
-{% endif %}
-
-{% if page.header.overlay_image contains "://" %}
-  {% capture overlay_img_path %}{{ page.header.overlay_image }}{% endcapture %}
-{% elsif page.header.overlay_image %}
-  {% capture overlay_img_path %}{{ page.header.overlay_image | relative_url }}{% endcapture %}
-{% endif %}
+{% capture img_path %}{{ page.header.image | relative_url }}{% endcapture %}
```
mmistakes: Think these captures can all be removed and just use the page variables directly. `{{ page.header.cta_url | relative_url }}`, `{{ page.header.overlay_image | relative_url }}`, etc.

iBug: For these three captures I removed the first two. The last one `overlay_img_path` is used multiple times later so I find it better to be kept.

mmistakes:
Nice work on this one, everything looks good to me!

Issue #2388, opened 2020-02-03
Masthead - get external links to open in new tab


projectgotheek:
There's this code in the masthead.html page:

{%- if link.url contains '://' -%}
              {%- assign url = link.url -%}

Is there a way to get this to open in a new tab, as in _blank

I've seen it done (though with a different implementation) in jekyll-documentation-theme, where they do it like this:

{% if item.external_url %}

                <li><a href="{{item.external_url}}" target="_blank" rel="noopener">{{item.title}}</a></li>

PR #2392, opened 2020-02-07
Masthead: Open absolute links in new tab

iBug:

## Summary

If a masthead link contains `://`, open it in a new tab via `target="_blank"`

## Context

#2388

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Enhancements
 
 - Bump Jekyll gem dependency to `v3.7`.
+- Open external masthead links in new tab. [#2392](https://github.com/mmistakes/minimal-mistakes/pull/2392)
 
 ### Bug Fixes
 
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -13,13 +13,12 @@
         </a>
         <ul class="visible-links">
           {%- for link in site.data.navigation.main -%}
-            {%- if link.url contains '://' -%}
-              {%- assign url = link.url -%}
-            {%- else -%}
-              {%- assign url = link.url | relative_url -%}
-            {%- endif -%}
             <li class="masthead__menu-item">
-              <a href="{{ url }}" {% if link.description %}title="{{ link.description }}"{% endif %}>{{ link.title }}</a>
+              <a href="{{ link.url | relative_url }}"
+                {% if link.url contains "://" %}target="_blank"{% endif %}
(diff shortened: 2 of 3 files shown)


mmistakes:
I've commented on other PRs that have been similar in nature.
I'm not a fan of forcing `target="_blank"` on the user for all the reasons cited in this post.

iBug:
@mmistakes #2445 sounds like a reasonable request - site creators should be able to determine open mode for each link. Perhaps this could be implemented?

mmistakes:
My view is 100% inline with those on the CSS Tricks article. It's easy enough to <kbd>CTRL</kbd> or right click a link to open in new window/tab. I'd rather the user decide when to do that instead of making that decision for them. It feels like an anti-pattern and I don't want to encourage it's use by adding a config for it in the theme.

If the site owner doesn't agree with that there's a mechanism to change it... override `_includes/masthead.html`.

> ## A Bad Reason: "Internal" links and "External" links are different.
>
> We'll have "internal" links (links that point to our own site) behave normally, but "external" links (links that point to other sites) open in a new window/tab.
>
> This is related to the above two reasons, only perhaps worse. You understand that normal style links are ideal, but if that means a user leaves your site, you're willing to break that ideal.
>
> I've heard from a lot of people that this is "a convention." As in the way it's supposed to be done. It's not.

There are valid reasons to use `target="_blank"`, opening external links that way doesn't feel like one of them.

iBug:
@mmistakes I'm sure there are more argumentative reasons to enable this change than staying at the status quo. Right-clicking requires an extra mouse action, while Ctrl-Click opens the new tab in the background (and requires more than just an extra click). What's more, opening in new tabs on mobile / touchscreen devices requires *significantly* more hold and delay, as one has to hold the link for the pop-up containing the "open in new tab" option (I'm not a fan of gestures - they're too much slower and more complex than simply tapping).

brackendawson:
I came here because an external site I link to is sometimes slow and users of my site were thinking my site is broken, had the idea to open in a new tab. For sure doing this maybe makes people spend more time on your site, but it does so with hostile design, and I want to avoid hostility, if someone is on my site it will be because they want or need to be. The linked CSS tricks article pulled me back from the brink.

In the end I settled on a more user-friendly approach, simply mark the link as external by prefixing its title with:
```html
<i class="fas fa-external-link-alt" aria-hidden="true"></i> My Amazing Link
```
Which puts a little "external" icon in front of it.

PR #2419, opened 2020-03-02
Share on Mastodon & Reddit button added

stafwag:

## Summary

* Share on Reddit and Mastodon buttons added.
* Added the possibility to enable/disable share on buttons in _config.yaml.
* Share on Twitter, Facebook and Linkedin is still the default.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -83,6 +83,12 @@ yandex_site_verification :
 naver_site_verification  :
 
 # Social Sharing
+share_buttons:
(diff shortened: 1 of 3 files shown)


stafwag:
I think that having the possibility to activate Mastodon and Reddit share buttons can be useful to a lot of users that use this great Jekyll theme.

If there is something that you don't like about the patch please let me know and I will try to fix it.

PR #2431, opened 2020-03-11
Use first_page_path from Paginate V2 if available

iBug:

Use `paginator.first_page_path` from Paginate V2 plugin for better compatibility. The original `jekyll-paginator` doesn't have this variable so this PR will not affect sites using original paginator.

I was reviewing my local overrides of the theme when I noticed this change and I even forgot what the problem was (what this change solved) but I hope this is helpful.


iBug:
I managed to recall what the problem was. Paginate V2 complains if `paginate_path` is set (not null) in `_config.yml` so that variable is unavailable.

Issue #2440, opened 2020-03-16
Author profile links not working on mobile


johannesnicolaus:
## Environment

On minimal mistakes website

## Expected behavior

The links when clicked should open the profile links.

## Steps to reproduce the behavior

Open the samples post.

https://mmistakes.github.io/minimal-mistakes/year-archive/

On mobile, when I click the follow button on the author, the dropdown appears, however when I click on the social media link it does not open the link but closes the dropdown.

![2020-03-16_22-06-12](https://user-images.githubusercontent.com/32991961/76761266-7c095180-67d2-11ea-99e4-8cd3e140c0de.gif)

valillon:
This issue seems to be unresolved, at least for iOS Safari.
![trim F749B7A4-A371-403B-92E6-1C471C26A82E](https://user-images.githubusercontent.com/2164412/137459875-7e30f714-fccc-4870-ad42-d5cf252fb2c5.gif)
.

PR #2441, opened 2020-03-16
Fix "Author profile links not working on mobile" (#2440)

iBug:
#2440


mmistakes:
Thanks for the PR. Noticed this after I pushed a fixed so going to close.

PR #2474, opened 2020-03-29
Added the possibility to avoid the recent posts heading in home

giuseppetotaro:

## Summary

The `home` page layout is built to show a list of recent posts after the **recent posts** heading (`<h3>`):

Screen Shot 2020-03-29 at 11 53 57 PM

However, it should be convenient to have the configuration option to disable the **recent posts** heading because home pages in many blogs usually don't have such heading and start directly with the content (i.e., blog posts).

This PR includes the new configuration parameter `recent_posts_title_off` in **_config.yml**, that is set to `false` by default (if it is set to `true`, the heading is not displayed), the conditional check in the `home` layout (**home.html**) and the updated documentation.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -27,6 +27,7 @@ repository               : # GitHub username/repo-name e.g. "mmistakes/minimal-m
 teaser                   : # path of fallback teaser image, e.g. "/assets/images/500x300.png"
 logo                     : # path of logo image to display in the masthead, e.g. "/assets/images/88x88.png"
 masthead_title           : # overrides the website title displayed in the masthead, use " " for no title
+recent_posts_title_off   : # true, false (default)
 # breadcrumbs            : false # true, false (default)
 words_per_minute         : 200
 comments:
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -4,7 +4,9 @@ layout: archive
 
 {{ content }}
 
+{% unless site.recent_posts_title_off %}
 <h3 class="archive__subtitle">{{ site.data.ui-text[site.locale].recent_posts | default: "Recent Posts" }}</h3>
+{% endunless %}
 
 {% if paginator %}
   {% assign posts = paginator.posts %}
--- a/docs/_docs/10-layouts.md
+++ b/docs/_docs/10-layouts.md
@@ -337,6 +337,12 @@ paginate_path: /blog/page:num
 **Note:** Jekyll can only paginate a single `index.html` file. If you'd like to paginate more pages (e.g. category indexes) you'll need the help of a custom plugin. For more pagination related settings check the [**Configuration**]({{ "/docs/configuration/#paginate" | relative_url }}) section.
 {: .notice--info}
 
+To avoid the "recent posts" heading, set to `true` the `recent_posts_title_off` in **_config.yml**.
+
+```yaml
+recent_posts_title_off: true
+```
+
 ## Splash page layout
 
 For full-width landing pages that need a little something extra add `layout: splash` to the YAML Front Matter.


ali-farooq0:
This feels too specific to one page, for these kind of changes, I've just made a copy of the layout file (in this case it'll be `_layout/home.html` and did the changes locally.

iBug:
This might be an overkill. I use the following CSS to get away with it:

```css
.layout--home .archive__subtitle {
  display: none;
}
```

ali-farooq0:
> This might be an overkill. I use the following CSS to get away with it:
> 
> ```css
> .layout--home .archive__subtitle {
>   display: none;
> }
> ```

Ahh yes much better 👍

giuseppetotaro:
Thanks @ali-farooq0 and @iBug for your quick reply. Really appreciated.

> This feels too specific to one page, for these kind of changes, I've just made a copy of the layout file (in this case it'll be `_layout/home.html`) and did the changes locally.

It is intended to be specific. The `home` layout is used only once in your website and it is for a very important page where you usually present your most important contents. This way theme's users would be able to easily set if they want the "recent posts" heading or not in `home` (many blogs don't report this heading or similar in home page).

> This might be an overkill. I use the following CSS to get away with it:
> ```
> .layout--home .archive__subtitle {
>   display: none;
> }
> ```

Using a CSS is a good idea, although `display: none` is [bad for accessibility](https://css-tricks.com/places-its-tempting-to-use-display-none-but-dont/). Indeed, I would prefer the @ali-farooq0's idea of creating a new layout, even though this would cause some redundancy in case my home is equal to the "original" `home` (besides the heading). 

Thanks again. Looking forward to your feedback.

iBug:
> Using a CSS is a good idea, although `display: none` is bad for accessibility.

I would argue that for this specific case it's actually a decent solution because all you want to do is to completely hide the "Recent Posts" heading from users. I see `display: none` doing no harm here.

> The `home` layout is used only once in your website

I'm actually using it *twice* ([1](https://github.com/iBug/iBug-source/blob/264e9e5/blog/index.html#L6), [2](https://github.com/iBug/iBug-source/blob/264e9e5/_pages/cn-index.md#L3)). It's all too common for people to list different types of posts, where for me it's classified by language.

giuseppetotaro:
Thanks again @iBug for your feedback.

> I would argue that for this specific case it's actually a decent solution because all you want to do is to completely hide the "Recent Posts" heading from users. I see `display: none` doing no harm here.

Fair enough.

> I'm actually using it _twice_ ([1](https://github.com/iBug/iBug-source/blob/264e9e5/blog/index.html#L6), [2](https://github.com/iBug/iBug-source/blob/264e9e5/_pages/cn-index.md#L3)). It's all too common for people to list different types of posts, where for me it's classified by language.

I agree with that. Indeed, I was referring to the specificity of the `home` page layout which should be, in my opinion, easily customizable via configuration such as putting on (or off) the "recent posts" heading.

I still think this configuration parameter would be a good idea to let users customize "a little bit more" via **_config.yml** what is generally used as the first page of their website/blog.
Apart from that, I am glad that I received from you all other ideas on how to hide the "recent posts" heading which could also be applied to other situations.

Thank you all very much.

iBug:
What if this is made a per-page configuration? An extra site-wide key just for hiding text that appears only once for most people seems like overkill. I would suggest that you go one step further, by letting users decide what text should appear there, rather than a simple ON/OFF toggle.

I couldn't find the reference, but I recall that the theme's owner @mmistakes once said he was not in favor of more configuration keys for added complexity.

mmistakes:
Thanks for the PR @giuseppetotaro, but as others have replied this sort of customization is what I'm trying to avoid. The `home` layout is small enough that to deviate from the what the original theme intended isn't that big of a deal to override it with a local version of `_layouts/home.html`.

Closing as there are mechanisms in place to make this sort of edit, and I don't think the need is large enough to offer a dedicated config flag to enable/disable it.

giuseppetotaro:
> What if this is made a per-page configuration? An extra site-wide key just for hiding text that appears only once for most people seems like overkill. I would suggest that you go one step further, by letting users decide what text should appear there, rather than a simple ON/OFF toggle.

I guess this is already possible if you modify the text in [ui-text.yml](https://github.com/mmistakes/minimal-mistakes/blob/master/_data/ui-text.yml) and the related translations. Instead, this site-wide key key could be easily used to enable/disable the heading in `home` which could be a very common situation, even if it is for one page (but there could be more as you wrote). 

> I couldn't find the reference, but I recall that the theme's owner @mmistakes once said he was not in favor of more configuration keys for added complexity.

Easy. You're talking about the [README](https://github.com/mmistakes/minimal-mistakes#contributing) of the repo. However, I don't think this is the case for the reasons I've explained in my previous comments. I agree with you that there are other good solutions such as using CSS (thanks for this suggestion), but this configuration would also be very useful and not overkill. It's just my opinion.

Thanks @iBug for your feedback.

giuseppetotaro:
> Thanks for the PR @giuseppetotaro, but as others have replied this sort of customization is what I'm trying to avoid. The `home` layout is small enough that to deviate from the what the original theme intended isn't that big of a deal to override it with a local version of `_layouts/home.html`.
> 
> Closing as there are mechanisms in place to make this sort of edit, and I don't think the need is large enough to offer a dedicated config flag to enable/disable it.

No worries. Thanks @mmistakes, @iBug, @akhy for your feedback.

PR #2516, opened 2020-05-04
Remove bold and italics to make it searchable

Anteus:

## Summary

Remove bold and italics that are noted with asterisks (*) to make it searchable with lunr and full text search.

### DIFF
--- a/assets/js/lunr/lunr-store.js
+++ b/assets/js/lunr/lunr-store.js
@@ -27,6 +27,7 @@ var store = [
               replace:"</h4>", " " |
               replace:"</h5>", " " |
               replace:"</h6>", " "|
+              replace:"*", " "|
             strip_html | strip_newlines | jsonify }},
           {%- else -%}
             {{ doc.content | newline_to_br |


iBug:
Is this really a problem? HTML is already stripped:

<https://github.com/f7770ad0799b1c847ea35972f1d286e71a1a50ee/minimal-mistakes/blob/master/assets/js/lunr/lunr-store.js#L30>

<https://github.com/f7770ad0799b1c847ea35972f1d286e71a1a50ee/minimal-mistakes/blob/master/assets/js/lunr/lunr-store.js#L41>

Anteus:
The search didn't work without that as the asterisks where around words still eg. *sample*

iBug:
I don't understand. I mean (properly placed) asterisks in **source Markdown** will be rendered into HTML tags `<b>` and `<em>` respectively, which is what *actually* makes text look bold and italic, so there should be no more raw asterisks in the output HTML, which is the script working on.

mmistakes:
It's unclear to me what problem this is solving. Can you link to a repo that is having issues with bold or italic words not being searchable?

As noted by @iBug HTML is already stripped. Stripping asterisks where you have in this PR will remove it from the text... it's not Markdown in this case.

Anteus:
My bad was using [this](https://github.com/solarillion/solarillion.github.io/blob/master/assets/js/lunr/lunr-store.js) modified one to index pages.

Issue #1428, opened 2017-12-30
Question: Is it possible to include the date below posts in home and single layouts?


dvhart:
- [x] This is a question about using the theme.
- [?] This is a feature request.
- [ ] I believe this to be a bug with the theme.
  - [x] I have updated all gems with `bundle update`.
  - [x] I have tested locally with `bundle exec jekyll build`.

---

## Environment informations

- **Minimal Mistakes version:** 4.8.1+
- **Jekyll version:** ?
- **GitHub Pages hosted:** yes
- **Operating system:** ?

---

## Expected behavior

First off - wow, great work. A highly configurable theme that actually works with gh-pages. A rare gem... ahem...

I would love to be able to see the post date below the title in the home listing as well as near the top (instead of the bottom) in the single layout. I didn't find a way to do this in the docs, but hoping I missed it!

## Steps to reproduce the behavior
Just visit any page:
https://dvhart.github.io

mmistakes:
Now way to do this via a config or setting. You’ll need to modify includes to reposition or add the date in different locations.

The home layout loops over the following include, so you’ll need to drop in the [date Liquid/HTML](https://github.com/mmistakes/minimal-mistakes/blob/master/_layouts/single.html) there.

https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/archive-single.html

dvhart:
Thanks @mmistakes. I started to modifications and got something mostly working. One of the goals here was to minimize any changes to the theme and to keep it separate from my repository with a gem. Hoping I can find something that is minimally invasive and configurable that you might consider including. I'll share whatever I come up with. Thank you for the response.

dvhart:
I've completed the overrides necessary to make this "work" for me. I've abused the page__meta class a bit to display both the date and the read-time. I'd like it to be smaller, and it doesn't work quite right with the archive tiles.

The home page now displays the date as well as the read-time. This could be smaller, and maybe gray or a lighter weight.
![screen shot 2017-12-31 at 1 04 55 am](https://user-images.githubusercontent.com/36253/34460566-e4cc30f8-edc6-11e7-94c0-85ec6044f7b5.png)

The same code displays now just above the content, regardless if overlay is used or not. This should definitely be smaller.
![screen shot 2017-12-31 at 1 05 05 am](https://user-images.githubusercontent.com/36253/34460572-205d8c2a-edc7-11e7-8370-2a15f8db6f18.png)

[...]

PR #2526, opened 2020-05-09
Show date of posts

limjh16:

## Summary

Adds an optional "date" alongside the read time in _includes/read_time.html

As a result of this change, I renamed read_time.html to post__meta.html

## Context

To enable the date, it is similar to enabling the read time, just set show_date to true in _config.yml

<details>
<summary>Date only screenshots</summary>

Screenshot 2020-05-09 at 10 22 58 PM
Screenshot 2020-05-09 at 10 22 56 PM
Screenshot 2020-05-09 at 10 22 53 PM

</details>
<details>
<summary>Date+read time screenshots</summary>

Screenshot 2020-05-09 at 10 23 25 PM
Screenshot 2020-05-09 at 10 23 23 PM
Screenshot 2020-05-09 at 10 23 27 PM
Screen Shot 2020-05-09 at 22 23 45

</details>


iBug:
Looks exceptionally awesome! I like it. Not sure what @mmistakes things though, as he's been resistant to adding extra configurations ([ref](https://github.com/mmistakes/minimal-mistakes/issues/1428#issuecomment-354612807)).

iBug:
For the file name, I suggest `post__meta.html` as it's clearer than "rtsd", which is pretty obscure per se.

limjh16:
> Looks exceptionally awesome! I like it. Not sure what @mmistakes thinks though, as he's been resistant to adding extra configurations ([ref](https://github.com/mmistakes/minimal-mistakes/issues/1428#issuecomment-354612807)).

Woops, should have done a check to see whether others have suggested this before... sorry about that

limjh16:
Hey, I had tested that before but in _includes/archive-single.html page.whatever doesn't work. Not sure why though...

iBug:
I would probably do something else like `{% assign page = post %}` before including the new file in the case you described. Or maybe `{% assign page = page | default: post %}` at the top of the new file. Differentiating between variables makes the code less maintainable.

mmistakes:
The complexity comes from `site.pages`, they don't have dates generally. This is the big difference between posts and pages/collections. Can't assume the user will add a `date` to the YAML front matter either which can really start to complicate your Liquid spaghetti code.

If I were to consider adding dates inline with read time I'd do it like my other themes e.g. [So Simple](https://github.com/mmistakes/so-simple-theme).

iBug:
@mmistakes It could be documented. Jekyll supports parsing date from the `yyyy-mm-dd-post-title.md` filename format even for documents other than those in the \_posts collection (e.g. custom collection). If the entire ecosystem around this theme's setup is properly documented, I'd say I see no problem implementing this - it's a vital piece of information for a post as iterated multiple times by multiple users in #1428.

mmistakes:
@iBug That's fine. To your point about reducing complexity. I'd suggest looking how I did it in my other themes, it's very similar to your suggestions. Basically normalize the variable names with assigns, then they can be piped through 1 include.

Simplifies the logic so you can do: If site dates are enabled and the document has a date set then show date. If read time is enable show that. Instead of if post... do this... if page do that like the original PR had.

limjh16:
doing `{% assign page = post | default page %}` seems to work, thanks for the suggestion @iBug!

_includes/post__meta.html
```diff
@@ -0,0 +1,29 @@
+{% assign page = post | default: page %}
```
iBug, line 1: I think stuff like `p="post"` can be dropped as well since you've already done this. There are 4 of them I can spot for now.

iBug:
I think it's better to put post date *before* read time for being more intuitive.

_includes/post__meta.html
limjh16: oops... missed those out, it's fixed now, thanks!

_includes/archive-single.html
```diff
@@ -24,11 +24,7 @@ <h2 class="archive__item-title" itemprop="headline">
         <a href="{{ post.url | relative_url }}" rel="permalink">{{ title }}</a>
       {% endif %}
     </h2>
-    {% if include.type == "grid" %}
-      {% include post__meta.html p="post" type="grid" %}
-    {% else %}
-      {% include post__meta.html p="post" %}
-    {% endif %}
+    {% include post__meta.html type = include.type %}
```
iBug, line 27: I haven't tested, but I'm afraid the spaces around the equal sign may lead to unexpected results. Also for the sake of consistency (nowhere else are there spaces like this in an `{% include %}` directive).

```diff
-    {% include post__meta.html type = include.type %}
+    {% include post__meta.html type=include.type %}
```

iBug:
@mmistakes Any ideas? This PR is at a perfect state now, even better than I could imagine. As discussed multiple times before, many users of the theme are in serious need of post dates on their websites. It's not hard to imagine that more users will happily enable it if it's easily available. I understand your view on "fewer configurations", but something so important like post date definitely worth it.

iBug:
Can we anticipate this fantastic feature in MM v4.20? 😊

thoemmi:
I'm waiting for it too, as I'd like to customize it to show tags as well.

mmistakes:
Finally got around to checking this out. Couple of small things then I think it's ready to merge in:

- [ ] Adjust spacing between datetime and read time. 

Screen Shot 2020-07-22 at 3 57 57 PM

The `|` doesn't have enough space between it and looks like it's part of the time. My suggestion would be to do something like `<span class="sep"></span>` instead. Then the following CSS which would allow someone to override the character and spacing easily without having to touch the `_include`.

```css
.sep::before {
  content: "\2022";
  padding-left: 0.5em;
  padding-right: 0.5em;
}
```

Screen Shot 2020-07-22 at 4 04 01 PM

- [ ] Add documentation about how to enable dates... perhaps before the [Reading time config here](https://mmistakes.github.io/minimal-mistakes/docs/configuration/#reading-time).

- [ ] Add a sample post with the date enabled similar to the [reading time posts](https://github.com/mmistakes/minimal-mistakes/blob/master/docs/_posts/2012-01-02-layout-read-time.md) to show an example and for testing purposes.

iBug:
To give more details on @mmistakes' sensible suggestions, you'll want to add the following changes:

- Replace your current separator (a plain vertical bar) with an empty span: `<span class="sep"></span>` and use the CSS `::before` pseudo-element to fill in the content and set the spacing. You'll want to add **mmistakes**' sample CSS in `_sass/minimal-mistakes/_page.scss`, under the "Page meta" section, probably right before the next section "Page taxonomy" (the blank line at L301), as shown below.

    <https://github.com/mmistakes/minimal-mistakes/blob/8a60d43b12d45b134613b1304ea5c46f76412bd8/_sass/minimal-mistakes/_page.scss#L277-L279>

- Add documentation in [`docs/_docs/05-configuration.md`][1]. You can compose the docs in a way similar to the "Read time" paragraph. Just follow the styles and you're good to go. Don't forget to take a screenshot.

    - You can even copy that paragraph and replace the content. IMO sentences, codes, etc. should remain the same, so just swap the keywords out and it'll be OK.

- Sample post: Copy the one **mmistake** linked and change the front matter. You may consider creating two copies, one with only post date enabled, the other with both.

  [1]: https://github.com/mmistakes/minimal-mistakes/blob/master/docs/_docs/05-configuration.md

limjh16:
Since the CSS "sep" class has already been used in [other places in the theme](https://github.com/mmistakes/minimal-mistakes/search?q=sep&unscoped_q=sep), I have opted to allow for customisation of the separator via ui-text.yml file, similar to the breadcrumbs_separator option.

To adjust spacing, just add spaces in the string, it will be changed to ```&nbsp;``` automatically in the html.

Is this okay?

iBug:
I would instead opt for another name, like `post__meta-sep` and stick to the CSS solution.

limjh16:
Any reason why? I would think editing the yml file would be easier, especially since the breadcrumb_separator option is in there too

iBug:
The YAML file can't be edited *in parts*, which means you have to copy the whole file in order to override a single key. Using `_config.yml` would be even better, though I still prefer the CSS solution.

limjh16:
Hmm, @mmistakes what do you think? It would be very simple to change it to use ```_config.yml``` and CSS anyways.

Also, I'm not really sure how you make the [preview images](https://mmistakes.github.io/minimal-mistakes/assets/images/mm-read-time-example.jpg), could you maybe help to make them? 😅

iBug:
A side note: It would be desirable to select-and-copy the full breadcrumb path including the separator, while it probably makes little sense to copy post date and reading time *along with* the separator. Using a CSS `::before` pseudo-element makes sure it's not selectable (which is how CSS-based Font Awesome icons work).

iBug:
@limjh16 If you have Jekyll locally installed (along with Ruby and Bundler), you can run `bundle exec jekyll serve` in the `docs` directory, which will open up a development server at <http://127.0.0.1:4000/minimal-mistakes/> for you to preview the whole docs site locally. You can then take the screenshots from your browser.

limjh16:
I was referring to the shadow + grey background, and actually to be more specific I do not currently have any image editing tools installed to add those effects (had to uninstall because I ran out of storage space woops)

iBug:
AFAICT that's the built-in screenshot utility in macOS. For this specific PR I think it's fine to take a screenshot with whatever you have - you're not trying to illustrate the whole page layout after all.

mmistakes:
I wouldn't worry about adding a screenshot. I think a description is clear enough what the feature is. Just as long as the documentation is there to explain how to toggle it on or off I'm cool.

It can always be improved later with a screenshot if needed.

iBug:
@mmistakes What about the implementation of the separator element? There have been 3 suggested in the thread:

1. The current approach, using text from `_data/ui-text.yml` as a text node
2. Similar to 1, but using text from `_config.yml` (probably with a default value) as a text node
3. Use an empty inline element (e.g. `<span>`) and apply CSS on its `::before` pseudo-element

My personal preference is 3 &gt; 2 &gt; 1, though any would be fine if it's easy enough to switch to 3 without overriding too many files (viva la CSS!).

mmistakes:
I prefer in the same order you proposed.

#3 since that's the method I suggest ;-) and it keeps the separator text out of the markup since it's decoration and not necessarily content.

#2 Is fine and matches how title separators are treated. Advantage over #3, it's "easier" to change by the user as they don't need to override any CSS which tends to be a show stopper for many. Disadvantage, the separator is in the HTML markup and there's more Liquid bloat introduced in the templates.

#1 Least desirable. Need to add a key for every language in the UI text data file versus a single key in `_config.yml` that's treated as a global variable.

limjh16:
I have changed the customisation to a CSS implementation

iBug:
@mmistakes This is looking perfect now (to me). Any chance you could take a look?

tobiasvl:
Thank you for this feature!

My page suddenly wouldn't build anymore (because I hadn't pinned a version number, it automatically used the latest version...), and after some digging I found out that the reason it broke was that `read_time.html` had been removed, which I used to – you guessed it – implement this exact functionality. Hehe.

👍

the-physicist:
I really liked it, you did a great job! Thank you.

Issue #2263, opened 2019-09-13
Sticky TOC pushes feature_row down


iBug:
## Environment

- Minimal Mistakes version: 4.16.6
- Ruby gem or remote theme version: Ruby gem
- Jekyll version: 3.8.6
- Git repository URL: Not required
- GitHub Pages hosted (if yes provide URL to site): Not required
- Operating system: Independent

## Wrong behavior

![image](https://user-images.githubusercontent.com/7273074/64869458-03370a80-d674-11e9-97da-ea8ee3baf685.png)

The sticky TOC pushes the first `feature_row` at an unexpected place. If the key `position: sticky;` was removed from the class `.sidebar__right.sticky`, then the `feature_row` locates at the correct place (i.e. no margin between it and the above heading 2 *Overview*).

I have since disabled `toc_sticky` for the article to give it a normal look.

## To reproduce

Create an article with a `feature_row` at top, as well as a few headings to give the TOC some height. Set `toc_sticky` on the page and check the result.

## Other

I'd like to know how to patch the CSS to fix this issue locally, but I'm not experienced in CSS and haven't yet figured out how. Yes, [I do patch the CSS](https://github.com/iBug/iBug-source/blob/master/assets/css/main.scss) to create my own desired layout.

mmistakes:
I this page hosted somewhere so I can inspect the HTML/CSS in browser?
Looks like the feature blocks are in a different `<div>` that comes after the one that has the main content and TOC.

Which is why you're seeing the space.

iBug:
@mmistakes [Here it is](https://ibugone.com/p/26-test).

mmistakes:
I don't know if this will have unintended consequences somewhere else but if you add `float: left` to `.feature__wrapper` it properly clear the sidebar.

```css
.feature__wrapper {
  float: left;
}
```

![Screen Shot 2019-09-14 at 11 07 54 AM](https://user-images.githubusercontent.com/1376749/64909949-eb956b80-d6df-11e9-9956-f9efa38ecf56.png)

iBug:
@mmistakes That's more a workaround than a solution to me. The feature block is now at a wrong position. The first block should be right under `<h2>Overview` without text between them but with `float: left;`, a few lines of text are moved up to fill in the gap.

What actually work for me was to add `.feature__wrapper { display: inline-block; }` and the order of content is well preserved. I don't know if there's any side effect either.

mmistakes:
[...]

PR #2540, opened 2020-05-21
Do not include feature rows in TOC

lsolesen:
This is a bug fix.

## Summary

The feature_row should not be included in the TOC

### DIFF
--- a/_includes/feature_row
+++ b/_includes/feature_row
@@ -21,7 +21,7 @@
 
         <div class="archive__item-body">
           {% if f.title %}
-            <h2 class="archive__item-title">{{ f.title }}</h2>
+            <h2 class="archive__item-title no_toc">{{ f.title }}</h2>
           {% endif %}
 
           {% if f.excerpt %}


iBug:
Good spot!

Just for the record, `feature_row`s aren't intended for posts ([ref](https://github.com/mmistakes/minimal-mistakes/issues/2263#issuecomment-531500264)):

> Not sure. Feature blocks weren't designed to be used as you have so that's likely why they misbehave for you.
> 
> They're more for splash pages, not posts with table of Contents.
> 
> Since you're doing images use the gallery helper instead. That's what it's designed for.
> 
> <https://mmistakes.github.io/minimal-mistakes/docs/helpers/#gallery>

lsolesen:
Feature rows works well for posts also. The gallery does not supply btn and
text, so they are also very helpful.

tor. 21. maj 2020 09.49 skrev iBug ♦ <notifications@github.com>:

mmistakes:
Would there be cases where someone would want the feature row headings to appear in the table of contents? If you're adding it as post/page content then I think it's valid to expect it should be reflected in the page hierarchy, especially if they are considered part of the "main" content.

There was an earlier PR to ignore the "Comments" section heading which made sense. Not entirely sure on this one. I can see users wanting them in the TOC. Which means some sort of config flag to enable/disable... which you all know my position on that one ;-)

lsolesen:
@mmistakes I think this should be kept open, or closed with a No Fix?

Issue #2541, opened 2020-05-21
Always adds whitespace to seo_description


lsolesen:
## Environment

- Minimal Mistakes version: 4.19.2
- Ruby gem or remote theme version:
- Jekyll version:
- Git repository URL: https://github.com/motionsplan/motionsplan-dk-jekyll
- GitHub Pages hosted (if yes provide URL to site): https://www.motionsplan.dk
- Operating system:

## Expected behavior

SEO description should not have an ending whitespace

## Steps to reproduce the behavior

seo_description consistently have a trailing whitespace.

iBug:
@mmistakes This one might be better categorized as a bug report.

PR #2542, opened 2020-05-21
Strip trailing whitespace in seo_description

lsolesen:
Related to #2541 . This will strip the extra whitespace in the end of seo_description.


iBug:
@mmistakes This one is simple enough. LGTM.

mmistakes:
@iBug looks that way, though I still like to do a sanity check. I've been bitten too many times from simple fixes.

PR #2550, opened 2020-06-02
UX Style Updates

davefinger:
UX Style Updates for all Pages


iBug:
This... incomprehensible...

PR #2575, opened 2020-06-20
Add `relative_url` filter to author home link

dianlujitao:
This is a bug fix.

## Summary

* Fix link for author name

## Context

No

### DIFF
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -17,7 +17,7 @@
 
   <div class="author__content">
     {% if author.home %}
-      <a href="{{ author_link }}"><h3 class="author__name" itemprop="name">{{ author.name }}</h3></a>
+      <a href="{{ author.home | relative_url }}"><h3 class="author__name" itemprop="name">{{ author.name }}</h3></a>
     {% else %}
       <h3 class="author__name" itemprop="name">{{ author.name }}</h3>
     {% endif %}


iBug:
意外捉住一只一加手机和 CyanogenMod 大佬！膜拜膜拜！

Please ignore the nonsense above.

---

The bug fix looks OK to me, but I'm negative towards whether we need the other feature. There's already the possibility to add custom scripts after the originals:

<https://github.com/mmistakes/minimal-mistakes/blob/bcd61266122892c2cdf3b5b85c12a71bf849d2fb/_includes/scripts.html#L25-L29>

It's as easy as putting your custom script as files somewhere and setting their paths in `_config.yml`, under the relevant key.

dianlujitao:
> The bug fix looks OK to me, but I'm negative towards whether we need the other feature. There's already the possibility to add custom scripts after the originals:
> 
> https://github.com/mmistakes/minimal-mistakes/blob/bcd61266122892c2cdf3b5b85c12a71bf849d2fb/_includes/scripts.html#L25-L29
> 
> It's as easy as putting your custom script as files somewhere and setting their paths in `_config.yml`, under the relevant key.

However, it's not feasible when we wanna use `<scripts defer>` and `<scripts async>`.

iBug:
That's a valid reason. I'm unsure about its usefulness unless you're loading large scripts where performance benefits from `async` and `defer` really matters. If you really need this, overriding a file [like it's done in the `/docs` site](https://github.com/mmistakes/minimal-mistakes/blob/master/docs/_layouts/default.html) should suffice.

dianlujitao:
> That's a valid reason. I'm unsure about its usefulness unless you're loading large scripts where performance benefits from `async` and `defer` stands out. If you really need this, overriding a file [like it's done in the `/docs` site](https://github.com/mmistakes/minimal-mistakes/blob/master/docs/_layouts/default.html) should suffice. Doing this on a per-site basis has always been the recommended way to further customize this theme.

I'm fine with that way, but an "official mechanism" looks better to me. Similar things are implemented for head and footer anyway.

mmistakes:
If you can update the PR to just have the author link bug fix I'll merge it in.
As far as the script enhancement, I'm working on something for another project and this is how I'm doing it there:

Liquid:

```
{% for script in site.footer_scripts %}
  {% unless script.inline %}
    <script src="{{ script.uri | relative_url }}" {{ script.attributes }}></script>
  {% else %}
    {{ script.inline }}
  {% endunless -%}
{% endfor %}
```

This allows for adding scripts as an array (how it's done today in the theme) or adding inline JavaScript. There is a new variable `attributes` for defining things like `async`, `crossorigin` etc.

Example:

```yaml
footer_scripts:
  - uri: https://code.jquery.com/jquery-3.4.1.min.js
    attributes: >
      integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo=" crossorigin="anonymous"
  - uri: /assets/js/my-custom-script.js
```

Would give you `<script src="https://code.jquery.com/jquery-3.4.1.min.js" integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo=" crossorigin="anonymous"></script>`

Inline script example:

```yaml
footer_scripts:
  inline: |
    <!-- Global site tag (gtag.js) - Google Analytics -->
    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-12345678-1"></script>
    <script>
      window.dataLayer = window.dataLayer || [];
      function gtag(){dataLayer.push(arguments);}
      gtag('js', new Date());

      gtag('config', 'UA-12345678-1');
    </script>
```

That said this would be a breaking change for the theme as `uri` would also have to be assigned.

mmistakes:
@dianlujitao Can you break these up into 2 PRs? I'd like to get the author fix in as that won't break anything. The script enhancement would have to be something we do later.

Issue #413, opened 2016-08-03
Author profile does not show links


rubinovk:
- [] This is a question about using the theme.
- [X] I believe this to be a bug with the theme --- not Jekyll, GitHub Pages or one of the bundled plugins.
- [ ] This is a feature request.
- [X] I have updated all gems with `bundle update`.
- [X] I have tested locally with `bundle exec jekyll build`.

---

## Environment informations

- **Minimal Mistakes version: v.3.3.6** 
- **`github-pages` or `jekyll` gem version: 89** 
- **Operating system: OSX** 

---

## Expected behavior

As per Configuration and https://mmistakes.github.io/minimal-mistakes/docs/layouts/#author-profile 

```markdown
author:
  name   : "Your Name"
  avatar : "bio-photo.jpg"  # placed in /images/
  bio    : "My awesome biography constrained to a sentence or two goes here."
  email  : # optional
  uri    : "http://your-site.com"
``` 
should make the author uri visible with an icon, as well as others, if filled in.

## Actual behavior

Non of the possible authorprofile links is visible. Neither in the https://mmistakes.github.io/minimal-mistakes/about/ example sidebar that should have author uri  `uri :"https://mademistakes.com"`

![screen shot 2016-08-03 at 14 13 30](https://cloud.githubusercontent.com/assets/2063487/17365108/82172a1c-5984-11e6-87b1-ec85c047b625.png)

Thanks for the great theme!

rubinovk:
After additional testing found that it's my browser's AdBlock that disables these links. Why would it? :$

mmistakes:
It's a known issue with adblockers and Font Awesome. The links are there, you're just not seeing them because your adblocker is removing them.

Here's the upstream issue if you want more context. https://github.com/FortAwesome/Font-Awesome/issues/1799

rubinovk:
Thanks for your quick reply and the context.

proof-by-accident:
Hi I'm actually having a similar problem- my social media links are showing up, but the FontAwesome icons are not.

_config.yml:
`author:                                                                                               
  name: proof-by-accident                                                                             
  avatar: "/images/prof_pic.jpg"                                                                      
  twitter: ProofByAccident                                                                            
  github: proof-by-accident                                                                           
  reddit: proof_by_accident  `

I added the Reddit link following the "Layout Customization" page, but no icons appeared.  Do I have to download the icon images and store them somewhere specific or something?

mmistakes:
[...]

PR #2583, opened 2020-06-28
Load Font Awesome CSS from jsDelivr CDN

iBug:

## Summary

Instead of using Font Awesome Kits, load its CSS from [jsDelivr CDN][1].

### Why?

✔ Font Awesome official source and NPM package
✔ Always up-to-date with latest FA 5 icon sets
✔ Faster everywhere (jsDelivr is the largest public NPM CDN, also works in China)
✔ Loads fewer bytes
✔ No bothering with `async` and/or `defer`. Always loaded asynchronously and displayed as soon as it's ready
✔ Works with ads blocker
✔ Works even with JavaScript disabled

## Context

Potentially related: #413

  [1]: https://www.jsdelivr.com/package/npm/@fortawesome/fontawesome-free

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Enhancements
+
+- Replace Font Awesome Kits with CSS from jsDelivr CDN. [#2583](https://github.com/mmistakes/minimal-mistakes/pull/2583)
+
 ## [4.19.3](https://github.com/mmistakes/minimal-mistakes/releases/tag/4.19.3)
 
 ### Enhancements
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -13,6 +13,7 @@
 
 <!-- For all browsers -->
 <link rel="stylesheet" href="{{ '/assets/css/main.css' | relative_url }}">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@5/css/all.min.css">
 
 <!--[if IE]>
   <style>
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -4,7 +4,6 @@
   {% endfor %}
 {% else %}
   <script src="{{ '/assets/js/main.min.js' | relative_url }}"></script>
-  <script src="https://kit.fontawesome.com/4eee35f757.js"></script>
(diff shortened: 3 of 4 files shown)


iBug:
If you dislike dependencies on external services, Font Awesome themselves provides CSS-based usage on

<https://use.fontawesome.com/releases/v5.13.1/css/all.css>

However I can't find a way to automatically track the latest releases in this way (jsDelivr can).

mmistakes:
This is why I moved to their newer kit feature so I didn't have to constantly bump the dependency since the version number was tied to URL.

The other consideration was I specifically went with their SVG solution as that was at the time was new and preferred installed method. Web fonts have sort of fallen out of favor, though I guess that might not hold much water since the project is called "FontAwesome" and still the primary way its used.

My understanding the JavaScript rewrites the `<i>` tags into inline SVG which may or may not have benefits. I honestly can't remember.

iBug:
I don't think they make any visual / practical difference, as even [the official FA docs](https://fontawesome.com/how-to-use/on-the-web/setup/using-package-managers) says (bottom of the page):

> **Reference either** the `/css/all.css` or the /js/all.js`, which contains everything you need ...

So for end users both methods of loading should be equivalent.

And yes, the JS version *does* rewrite the `<i>` tags, which isn't per-se bad. So what about improving it one step further with this new approach?

mmistakes:
I didn't think it made any visual differences, it was more of a use SVG or webfonts for icons debate and if one was more performant, accessible, etc. Since the SVGs aren't getting inlined into the source HTML I don't think it matters much, so I'm fine with going back to the webfont/CSS solution for icons.

iBug:
As a follow-up: I just noticed that FA's official kits also have a "latest" reference, for example:

```html
<link rel="stylesheet" href="https://kit-free.fontawesome.com/releases/latest/css/free.min.css">
<!-- or -->
<script type="text/javascript" href="https://kit-free.fontawesome.com/releases/latest/js/free.min.js">
```

FA's official CDN is based on [StackPath](https://www.stackpath.com/products/cdn/), which may be faster or slower. As far as I am concerned, StackPath is also one of the supporters of jsDelivr, so logically jsDelivr shouldn't be slower or less stable than SP.

PR #2608, opened 2020-07-18
[m]: Updated liquid tag version

hbvj99:

## Summary

I have updated liquid tags version 3 to 4 this will help on removing occasional whitespaces from HTML document.

## Context

There were occasional whitespaces generated by liquid tag version 3. Liquid [doc](https://shopify.github.io/liquid/basics/whitespace/) describes removing unnecessary spaces generated by ```_includes``` ```_layouts``` by simply updating a hyphen in tag syntax. Github [dependencies](https://pages.github.com/versions/) are already up to date with the new version 4.


iBug:
This is an interesting change. I wonder if what we have in the status quo *really* is a problem. I doubt many people care about this.

hbvj99:
@iBug I agree that most people wouldn't care, but removing extra unnecessary whitespaces wouldn't hurt. It will help to decrease page overall size but noticeable difference is small I'd say.

iBug:
I don't think so. It provides minimum benefit, but adding `{%-` and `-%}` everywhere makes the source code less maintainable. If you're really looking for cleaner HTML output, you should probably look into plugins like [jekyll-tidy][1] or such.

I could be wrong, so let @mmistakes decide.

  [1]: https://github.com/apsislabs/jekyll-tidy

mmistakes:
My thinking is inline with @iBug’s.

Where collapsing Liquid white space makes sense I think is in some of the `for` loops as those can introduce a lot of blank lines. But placing on every bit of Liquid code is overkill when minifying after the fact thru a build process or plugin is an option.

I haven’t looked at the HTML your PR produces yet, but in my experience placing `-` on the start and end of some tags will eat line breaks making HTML harder to read. 

I know some users of the theme aim for a readable / “beautified” HTML source even though minifying is better for performance to reduce file size.

hbvj99:
Sure thing @mmistakes, I thought this could be useful. I'll close this PR for now :)

PR #2613, opened 2020-07-23
Allow adding extra attributes to scripts

dianlujitao:

## Summary

site.footer_scripts is not enough in case we wanna defer or
asynchronously load custom scripts

## Context

https://github.com/mmistakes/minimal-mistakes/pull/2575


iBug:
I wonder if there's a way to keep the compatibility with the status quo. This is a breaking change so it'd be better if there's a "transition period".

Also, your implementation of inlined `site.footer_scripts` doesn't match that of your updated documentation. With

```liquid
{% if site.footer_scripts %}
  {% for script in site.footer_scripts %}
    {% unless script.inline %}
```

The config should have been like this:

```yaml
footer_scripts:
  - inline: "whatever"
```

Perhaps you wanted it like this:

```liquid
{% if site.footer_scripts %}
  {% if site.footer_scripts.inline %}
  {% else %}
    {% for script in site.footer_scripts %}
```

PR #2620, opened 2020-07-25
Exclude vendor / plugins JS from packed gems

iBug:
Excludes `assets/js/_main.js`, `assets/js/plugins/` and `assets/js/vendor/` from the packaged Ruby Gem since end users don't need them, as they're already packaged into `main.min.js` (which is kept, of course).


mmistakes:
Not sure if this has any benefit. Keeping all of the scripts in the bundled gem is useful for those who override the `site` footer scripts to disable or enable specific vendor scripts. 

It’s mentioned in the docs too as you’ll need to include all the vendor scripts if you use that feature. And if we exclude them from the gem the user needs to manually copy them over into their repo instead of just using those already bundled. 

https://mmistakes.github.io/minimal-mistakes/docs/javascript/#customizing

iBug:
Oops, missed that. Thanks for pointing out!

PR #2630, opened 2020-07-29
Add support for raw `background` CSS for page header overlay

iBug:

## Summary

Allow using custom raw CSS `background` value for page overlay like this:

```yaml
header:
  overlay_background: 'lightblue url("img_tree.gif") no-repeat fixed center'
```

Specifying this key overrides `overlay_color` and `overlay_image` since this is for advanced users with CSS knowledge.

### Note

The reason I chose `style=''` instead of `style=""` is because YAML performs no escaping for single-quote-enclosed strings, which would better match the common expectation that *"I'm using double quotes inside here (the YAML value) and it should also work over there"*.

### Changed

- [x] Core code for the functionality
- [x] Updated documentation for the new feature
- [x] Added an example post to show how it performs
- [x] Updated CHANGELOG and history

### Preview (new example post)

preview


iBug:
Struggled through debugging (it didn't work at all at first) and resorted to splitting quite a few other files. There were too many duplicated code.

Here's the summary:

- Split the include of page hero partial and breadcrumb partial into `page__header-common.html`, which itself includes a safeguard for not including breadcrumb partial in splash layout
- The `page__hero-common.html` partial also produces a `has_overlay` variable that can be used thereafter, even in other files that include it. This variable is now used to control many other things, like:
  - whether the page hero class has the `--overlay` suffix
  - whether page header is displayed *in* the hero `<div>`
  - the existence of the "inline header" when a page hero overlay is absent

mmistakes:
I'm on board with DRYing up the header/breadcrumb includes into a common one that is used across layouts.

But not sure on the CSS control via YAML Front Matter. I feel like it's introducing a bad pattern for future enhancements. I get the utility of the feature but the cost of introducing it might be high. What's to say there is a need to modify other CSS properties of the overlay other than `background`? Then we can expect enhancements to replicate other CSS properties in YAML... which isn't ideal.

I'm working on a new theme/framework and trying to think through these sorts of things. For example I'm doing something like this on the `body` element

```html
<body class="layout__{{ page.layout | default: layout.layout }}{% if page.classes or layout.classes %}{{ page.classes | default: layout.classes | join: ' ' | prepend: ' ' }}{% endif %} {{ page.title | slugify | prepend: 'page__' }}">
```

MM does something similar, but only injects the layout name as a class for additional styling hooks. The above takes it a step further by allowing you to add custom classNames at the `page` level (via YAML Front Matter) as well as slugifying the page's title and making that a className too.

Which would allow you to do your own styling for a specific page because you now have a CSS hook to do so:

1. `layout__layout-name`
2. `page__slugified-page-title`
3. Custom class name of your choosing

I get that you'd still need to know the className and make that association vs. dropping in CSS into a YAML variable. But this way seems more powerful.

```
---
title: My awesome post
---

<style>
  .my-awesome-post {
    color: red;
  }
</style>

Wow wee I have an awesome post that can be custom styled...
```

mmistakes:
Also, I wasn't thrilled with how overlay colors and such were originally built. This is another example of me being lax with a feature add and then that idea is picked up and built on as you've done in this PR. The idea is good, but the execution could benefit from a rethink.

So many other features in this theme are that way...

iBug:
I'm fine with defining CSS in CSS files, but the problem is, if I specify neither `overlay_image` nor `overlay_color`, the parent `<div>` doesn't have the overlay class. However if I specify either and proceed to customizing through CSS, I have to use `!important` which I generally dislike for reasons [all around online][1]. While this patch seems too specific for a single CSS key, it opens up the opportunity to get an overlay background **without having to `!important`**. That's what I was thinking at first.

  [1]: https://uxengineer.com/css-specificity-avoid-important-css/

mmistakes:
And that's the heart of the problem I have with the overlay header in general. The way it was built wasn't forward thinking and became a feature I let slide into the repo... and now it has technical debt to work through.

Maybe each header variant should have been their own classes instead of trying to do the different flavors of headers in the one... one with an image, another with text overlaid an image, another with a transparent background, yet another with a video, etc. etc.

Can fix some of these things with incremental enhancements like you've done here, but I feel like that's just a detour to the main event. Where I burn the whole thing down with a true 5.0 release.

Larger problem is this project is too many things to too many people and it's hard reconciling the gap between. At this point I almost feel like it needs to either be simplified greatly. And start a fork that becomes a framework/starter for the vocal minority who want extreme customization and flexibility.

mmistakes:
And for something named "Minimal Mistakes" I sure see a lot of mistakes as evident from the issues submitted daily to this repo. And most of that is the theme's fault because it has too many options to weed through and setup before you get a half decent looking site.

iBug:
That's true. It's overcomplicated at present, to be honest. The idea of a great refactor (release 5.0) sounds intriguing. I'd be glad to give a hand should you decide to proceed.

mmistakes:
Will do, and I'll certainly loop you in if/when a 5.0 version starts to materialize. You've been very helpful fielding issues I don't get to and submitting well put together PRs. I appreciate it!

PR #2640, opened 2020-08-04
Wrap up post date and read time in <span> with class

iBug:

## Summary

Post date and read time are now contained in an extra layer of `<span>` with BEM-named class, allowing for CSS-based customization (e.g. use different fonts for date and read time, and for post header and listing page).


iBug:
This should probably better handled in #2641.

PR #2641, opened 2020-08-04
Refactor page meta

mmistakes:
This is a bug fix.

## Summary

- [x] Fixes `page.` variables that were getting clobbered by an over zealous `assign`
- [x] Adds archive [`/test` pages](https://github.com/mmistakes/minimal-mistakes/tree/master/test/_pages) with grid views
- [x] Simplifies Liquid and removes extra `<br />` added for spacing reasons and replaces with CSS to achieve the same thing.

## Context

Fixes #2616


mmistakes:
@iBug and @TarodBOFH if you could give a once over and maybe test to make sure I didn't break anything else and this works as intended, would appreciate it. :smile:

TarodBOFH:
> @iBug and @TarodBOFH if you could give a once over and maybe test to make sure I didn't break anything else and this works as intended, would appreciate it. 😄

Sure; I know that I can reference a commit on the remote theme but, can I also reference a branch plus commit?

mmistakes:
@TarodBOFH I think you can. Check the remote theme docs https://github.com/benbalter/jekyll-remote-theme#declaring-your-theme

TarodBOFH:
> @TarodBOFH I think you can. Check the remote theme docs https://github.com/benbalter/jekyll-remote-theme#declaring-your-theme

Yup, I could. 
```
remote_theme: mmistakes/minimal-mistakes@b0b886f3515a5cdf138675f043f47eaa45a0696d
```

This references the last commit on this branch

I ran that on my docker image with the previous blog status (that is only this version of minimal-mistakes change) and so far it's looking good to me (site review, not code review)

_includes/page__meta.html
```diff
@@ -0,0 +1,26 @@
+{% assign document = post | default: page %}
```
TarodBOFH, line 1: document is still generic, I always like self contained variables, what about page_meta as variable?

_includes/page__meta.html
```diff
@@ -0,0 +1,26 @@
...
+      {% if words < words_per_minute %}
+        {{ site.data.ui-text[site.locale].less_than | default: "less than" }} 1 {{ site.data.ui-text[site.locale].minute_read | default: "minute read" }}
+      {% elsif words == words_per_minute %}
+        1 {{ site.data.ui-text[site.locale].minute_read | default: "minute read" }}
+      {% else %}
+        {{ words | divided_by:words_per_minute }} {{ site.data.ui-text[site.locale].minute_read | default: "minute read" }}
+      {% endif %}
+    {% endif %}
+  </p>
+{% endif %}
```
TarodBOFH, line 26: Just cosmetic:

Missing new line at EOF

_includes/page__meta.html
TarodBOFH: > document is still generic, I always like self contained variables, what about page_meta as variable?

Nevermind, shouldn't comment at 1am. good work. Document is the current `document` so it's good for me;

iBug:
I still find the introduction of an extra variable `document` unnecessary - using `post = post | default: page` is harmless and potentially less confusing. `post` isn't an "automatic" variable so there's no concern of "accidentally overwritten" with it.

mmistakes:
@iBug Semantically it's a bit weird using `post`. These are used for collection documents as well. I see a lot of Jekyll users stumbling over the whole `post` and `page` thing thinking it's some sort of automatic variable, which is why I lean more on using something like `document`.

TarodBOFH:
TBH I was inclined to use a defensive approach, like having a init _section_ on top of the file with all the needed variables, much like `{% assign date = document.date %}` but without the need of document, i.e. ` {% assign page_meta.date = page.date | default: post.date %}` and so on but I don't know the idiomatic way of doing that in liquid.

If other snippets also use `document` variable (not the case today), unexpected things might happen again. I was looking for some kind of standardisation here on liquid docs but couldn't find anything.

Just my two cents for future developments if applicable.

iBug:
Can you wrap post date and time in their own `<span>`s so it's possible to customize their looks by CSS (e.g. when one wants different looks for date and time)?

PR #2697, opened 2020-09-18
Fix Closing Tag of Figures Without Captions in Ordered and Unordered Lists

Leo3418:
This is a bug fix.

## Summary

This patch removes redundant `</figure>` tag rendered as text when the figure helper is used in a list, either ordered or unordered, and no caption is specified for that figure.

## Context

```md
---
layout: single
---

1. Lorem ipsum dolor sit amet

   {% include figure image_path="https://raw.githubusercontent.com/mmistakes/minimal-mistakes/master/screenshot.png" alt="Test image" %}

2. The quick brown fox jumps over the lazy dog
```

The above post is going to be rendered as in the image below. The figure is not captioned, which produces a redundant `</figure>` below it. Another noticeable issue is the next item that is supposed to be in the same list is put in a new list, resetting the number to 1.

localhost_4000_2020_09_18_original-figure-helper html

After this patch is applied, the same post can be rendered correctly.

localhost_4000_2020_09_18_patched-figure-helper html

The figure caption still works when the user specifies one.

```liquid
{% include figure image_path="https://raw.githubusercontent.com/mmistakes/minimal-mistakes/master/screenshot.png" alt="Test image" caption="hello, world"%}
```

localhost_4000_2020_09_18_patched-figure-helper-test html

### DIFF
--- a/_includes/figure
+++ b/_includes/figure
@@ -1,9 +1,9 @@
 <figure class="{{ include.class }}">
   <img src="{{ include.image_path | relative_url }}"
        alt="{% if include.alt %}{{ include.alt }}{% endif %}">
-  {% if include.caption %}
+  {%- if include.caption -%}
     <figcaption>
       {{ include.caption | markdownify | remove: "<p>" | remove: "</p>" }}
     </figcaption>
-  {% endif %}
+  {%- endif -%}
 </figure>


iBug (approved):
Nice catch!

mmistakes:
Just to clarify, the bug is Markdown related and not the logic in the `figure` include, correct?

Looks like all the fix is doing is collapsing some white space to remove the indent which Markdown interprets as a `<code>` element which is why it's not rendering the closing `</figure>`.

Leo3418:
> Just to clarify, the bug is Markdown related and not the logic in the `figure` include, correct?
> 
> Looks like all the fix is doing is collapsing some white space to remove the indent which Markdown interprets as a `<code>` element which is why it's not rendering the closing `</figure>` correctly.

Yes, that `figure` file's logic looks good to me. If it is used in normal text, not in any lists written in Markdown syntax, it actually works as intended.

PR #2723, opened 2020-10-30
Allow custom sorting for collections

n-elie:

## Summary

This PR allows sorting of collections by custom keys. This is done by using directly the `include.sort_by` value in sort tag instead of testing it's value.

### DIFF
--- a/_includes/documents-collection.html
+++ b/_includes/documents-collection.html
@@ -1,16 +1,14 @@
 {% assign entries = site[include.collection] %}
 
-{% if include.sort_by == 'title' %}
+{% if include.sort_by %}
   {% if include.sort_order == 'reverse' %}
-    {% assign entries = entries | sort: 'title' | reverse %}
+    {% assign entries = entries | sort: include.sort_by | reverse %}
   {% else %}
-    {% assign entries = entries | sort: 'title' %}
+    {% assign entries = entries | sort: include.sort_by %}
   {% endif %}
-{% elsif include.sort_by == 'date' %}
+{% else %}
   {% if include.sort_order == 'reverse' %}
-    {% assign entries = entries | sort: 'date' | reverse %}
-  {% else %}
-    {% assign entries = entries | sort: 'date' %}
+    {% assign entries = entries | reverse %}
   {% endif %}
 {% endif %}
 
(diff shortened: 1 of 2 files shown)


iBug:
Can you amend the documentation as well? It's all good to go then.

You may want to take a look at [`/docs/_docs/10-layouts.md`](https://github.com/mmistakes/minimal-mistakes/blob/master/docs/_docs/10-layouts.md).

_includes/documents-collection.html
```diff
@@ -1,16 +1,14 @@
...
   {% if include.sort_order == 'reverse' %}
-    {% assign entries = entries | sort: 'title' | reverse %}
+    {% assign entries = entries | sort: include.sort_by | reverse %}
   {% else %}
-    {% assign entries = entries | sort: 'title' %}
+    {% assign entries = entries | sort: include.sort_by %}
   {% endif %}
-{% elsif include.sort_by == 'date' %}
+{% else %}
   {% if include.sort_order == 'reverse' %}
```
iBug, line 10: Looks like we're doing repeated work here - we only need to check for `sort_order` once, after the `sort_by` check. Would you help fix this as well?

iBug (changes requested):

_includes/documents-collection.html
n-elie: Don't we need to be able to use the `sort_by` parameter even if `sort_order` is not specified?

iBug: To be exact, here's what we have for now:

```javascript
if (sort_by is given) {
  if (reverse) {
    entries = entries | sort_by | reverse;
  } else {
    entries = entries | sort_by;
  }
} else {
  if (reverse) {
    entries = entries | reverse;
  }
}
```

And here's what I'm expecting:

```javascript
if (sort_by is given) {
  entries = entries | sort_by;
}

if (reverse) {
  entries = entries | reverse;
}
```

Clearly the latter looks more concise, right?

n-elie: You're right, I have made the changes

iBug (approved):
@mmistakes All good to go for now. I'm in strong favor of this. Would you mind taking a look?

PR #2732, opened 2020-11-08
Fix show_excerpts in collection layout's front matter

CrazyOrr:
This is a bug fix.
## Summary
The `show_excerpts` front matter in [collection layout](https://mmistakes.github.io/minimal-mistakes/docs/layouts/#layout-collection) doesn't work as documented for now. Fixed it.

### DIFF
--- a/_includes/archive-single.html
+++ b/_includes/archive-single.html
@@ -10,6 +10,12 @@
   {% assign title = post.title %}
 {% endif %}
 
+{% if show_excerpts == false %}
+  {% assign show_excerpt = false %}
+{% else %}
+  {% assign show_excerpt = true %}
+{% endif %}
+
 <div class="{{ include.type | default: 'list' }}__item">
   <article class="archive__item" itemscope itemtype="https://schema.org/CreativeWork">
     {% if include.type == "grid" and teaser %}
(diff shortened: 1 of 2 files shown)


iBug:
not stale

_includes/archive-single.html
```diff
@@ -10,6 +10,12 @@
   {% assign title = post.title %}
 {% endif %}
 
+{% if show_excerpts == false %}
+  {% assign show_excerpt = false %}
```
iBug, line 14: Can we be consistent with plural form here? I'd go with `show_excerpts`. Same for 2 more occurrences below.

iBug (changes requested):

_includes/archive-single.html
CrazyOrr: Sure, as long as you don't mind overshadowing the original variable `show_excerpts`.

iBug: Hmmm, looks like there's no need for `show_excerpt` at all. Just use `show_excerpts != false` below and it'll be fine.

To be honest, I'm kind of puzzled about all these variables now.

CrazyOrr: Yes, good idea.

PR #2734, opened 2020-11-09
Update remote theme instructions

susanstevens:

## Summary

Update remote theme installation instructions to indicate that `jekyll-include-cache` should be added to the Gemfile.

## Context

I followed the [remote theme installation](https://github.com/mmistakes/minimal-mistakes#remote-theme-method) instructions and got the following error when running `bundle exec jekyll serve`:

```
Dependency Error: Yikes! It looks like you don't have jekyll-include-cache or one of its dependencies installed. In order to use Jekyll as currently configured, you'll need to install this gem. The full error message from Ruby is: 'cannot load such file -- jekyll-include-cache' If you run into trouble, you can find helpful resources at https://jekyllrb.com/help/! 
```

Adding the `jekyll-include-cache` gem to my Gemfile resolved the error, so it should probably be part of the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ To install:
    source "https://rubygems.org"
 
    gem "github-pages", group: :jekyll_plugins
+   gem "jekyll-include-cache", group: :jekyll_plugins
    ```
 
 2. Add `jekyll-include-cache` to the `plugins` array of your `_config.yml`.
--- a/docs/_docs/01-quick-start-guide.md
+++ b/docs/_docs/01-quick-start-guide.md
@@ -64,6 +64,7 @@ To install as a remote theme:
    source "https://rubygems.org"
 
    gem "github-pages", group: :jekyll_plugins
+   gem "jekyll-include-cache", group: :jekyll_plugins
    ```
 
 2. Add `jekyll-include-cache` to the `plugins` array of your `_config.yml`.


iBug:
This may not be necessary as the next point says:

> 2. Add `jekyll-include-cache` to the `plugins` array of your `_config.yml`.

susanstevens:
@iBug I thought calling out that you need to add `jekyll-include-gem` to the Gemfile would be helpful since step 2 only mentions adding it to `_config.yml`. I had to add it the Gemfile _and_ add it to the plugins array in `_config.yml`.

iBug:
Shouldn't be like that... The gem is included with `github-pages` and adding it in `_config.yml` should suffice.

akhyarrh:
> Shouldn't be like that... The gem is included with `github-pages` and adding it in `_config.yml` should suffice.

AFAIK, `jekyll-include-cache` is not included in `github-pages` gem: https://pages.github.com/versions/

`jekyll-include-cache` are added by theme: https://github.com/mmistakes/minimal-mistakes/blob/master/minimal-mistakes-jekyll.gemspec#L23

But maybe it will not loaded if theme is used via `remote_theme`, so that why Jekyll cant find `jekyll-include-cache` ?

mmistakes:
I believe @susanstevens and @akhyarrh are correct. `jekyll-include-cache` isn't a dependency of the GitHub Pages gem and needs to be included in the `Gemfile` to work.

mmistakes (approved):

mmistakes (approved):

PR #2735, opened 2020-11-10
WIP Use page.lang or site.lang if it exist

akhyarrh:

WIP. DO NOT MERGE

TODO:
- [ ] Update doc about this change

## Summary 
Use `page.lang` or `site.lang` if it exist on page front matter or site config.
`jekyll-feed` use site & post lang to generate `xml:lang` property on `feed.xml`.

## Reference
- https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml#L5
- https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml#L56

## Use case
User can use specific `lang` frontmatter if they want specific post or page to use different language.
Preserve fallback by still use `site.locale` if `page.lang` or `site.lang` dont exist.
Reason for using `lang` is because they are used by `jekyll-feed`.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -8,7 +8,16 @@
   Free for personal and commercial use under the MIT license
   https://github.com/mmistakes/minimal-mistakes/blob/master/LICENSE
 -->
-<html lang="{{ site.locale | slice: 0,2 | default: "en" }}" class="no-js">
+{% capture locale %}
+  {% if page.lang %}
+    {{ page.lang }}
+  {% elsif site.lang %}
+    {{ site.lang }}
+  {% else %}
+    {{ site.locale | slice: 0,2 }}
+{% endcapture %}
+
+<html lang="{{ locale | default: "en" }}" class="no-js">
   <head>
     {% include head.html %}
     {% include head/custom.html %}


_layouts/default.html
```diff
@@ -8,7 +8,16 @@
...
+{% capture locale %}
+  {% if page.lang %}
+    {{ page.lang }}
+  {% elsif site.lang %}
+    {{ site.lang }}
+  {% else %}
+    {{ site.locale | slice: 0,2 }}
+{% endcapture %}
+
+<html lang="{{ locale | default: "en" }}" class="no-js">
```
iBug, line 20: Using `{% capture %}` is going to be problematic here: It always assigns a string to the "capture variable", while the `default` is assigned for [only `false`, `null` and empty strings][1]. Since your capture contains newlines, the default never applies.

As there's no special string conversion involved (e.g. object → string), I'd suggest that you stick to `{% assign %}` even though [it may sometimes be slower than plain `{% capture %}`s][2], which I believe doesn't matter here.

  [1]: https://shopify.github.io/liquid/filters/default/
  [2]: https://github.com/Shopify/liquid/issues/487

_layouts/default.html
```diff
@@ -8,7 +8,16 @@
...
 -->
-<html lang="{{ site.locale | slice: 0,2 | default: "en" }}" class="no-js">
+{% capture locale %}
+  {% if page.lang %}
+    {{ page.lang }}
+  {% elsif site.lang %}
+    {{ site.lang }}
+  {% else %}
+    {{ site.locale | slice: 0,2 }}
+{% endcapture %}
```
iBug, line 18: Pretty sure there should have been an `{% endif %}` before this line.

iBug (changes requested):

Issue #2597, opened 2020-07-10
Cannot get fixurl.js (in 404.html)


floatingpurr:
## Environment

- Minimal Mistakes version: 4.19.3
- Ruby gem or remote theme version: n/a
- Jekyll version: 4.1.0
- Git repository URL: n/a
- GitHub Pages hosted (if yes provide URL to site): n/a
- Operating system: Docker based installation

## Expected behavior
The page [linkhelp.clients.google.com/tbproxy/lh/wm/fixurl.js](https://linkhelp.clients.google.com/tbproxy/lh/wm/fixurl.js) si not reachable anymore

## Steps to reproduce the behavior

Click [here](https://linkhelp.clients.google.com/tbproxy/lh/wm/fixurl.js)

## Other

This library is used in the 404.html page but it is not available.

mmistakes:
Looks like Google removed it. Not sure if there is a replacement or not.

mmistakes:
Looking like Google killed this feature.

![Screen Shot 2020-07-10 at 12 04 28 PM](https://user-images.githubusercontent.com/1376749/87174708-96959800-c2a5-11ea-8ed4-6e1cba442292.png)

https://webmasters.googleblog.com/2008/08/make-your-404-pages-more-useful.html

If someone wants to submit a PR to remove the script and text referencing "searcing" from [`404.md`](https://github.com/mmistakes/minimal-mistakes/blob/master/docs/_pages/404.md)...

PR #2737, opened 2020-11-13
Update 404.md

Jip-Hop:
This is a bug fix.

## Summary

Removed Google Search script which no longer worked.

## Context

https://github.com/mmistakes/minimal-mistakes/issues/2597


akhyarrh:
This is just a suggestion, but why dont we use theme's search feature ? Add a button that open search modal, or add a form to 404 page ?

PR #2756, opened 2020-12-20
Use sort_natural instead of custom-logic

gamue:

## Summary

- The sorting logic for tags and categories is done by custom-logic. I thought it's cleaner code to use `sort_natural` which exists out of the box for that use-case. See documentation at https://shopify.github.io/liquid/filters/sort_natural/
- Afaik this method was introduced in Jekyll 3.5, means it would be a requirement to have at least that version. However it's been released some years ago and is supported by GitHub Pages, so don't think it's breaking something

### DIFF
--- a/_includes/category-list.html
+++ b/_includes/category-list.html
@@ -6,19 +6,12 @@
 {% endcase %}
 
 {% if site.category_archive.path %}
-  {% comment %}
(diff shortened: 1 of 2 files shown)


iBug (approved):
LGTM

PR #2773, opened 2021-01-01
add extension point to add scripts at the end of html page

SebastienAndreo:

## Summary

I have created an overridable file `myscripts.html` and include it at the end of `default.html` to be able to control more accurately some JS execution. I also made the customization point more visible by creating a folder `custom` in the `_includes` folder.

## Context

It is sometimes useful to control the execution of some javascript being able to add them at
the end helped me to have the right rendering.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -37,6 +37,7 @@
     </div>
 
     {% include scripts.html %}
+    {% include custom/myscripts.html %}
 
   </body>
 </html>


iBug:
Consider taking a look at #2116 and #2613 first. I think it's the recommended way to override `scripts.html` directly, but I could be wrong.

tsmanner:
This sounds like a use for the [after_footer_scripts](https://mmistakes.github.io/minimal-mistakes/docs/javascript/#customizing) config option.  They're added [at the end](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/scripts.html#L24) of `scripts.html`, which I think is equivalent to what this PR does, if the intent is to add `<script src="/path/to/some/script.js"></script>` tags to the site.

PR #2774, opened 2021-01-02
feat: Search icon in masthead is a Font Awesome icon.

rhwood:

## Summary

Allow a Font Awesome icon to be used as the search icon instead of the built-in SVG.

## Context

This is not related to any existing issues, but is implemented as optional to:
1. not change the existing default behavior
2. allow other implementations of this site that are avoiding Font Awesome for GDPR and CCPA to continue to do so

### DIFF
--- a/_includes/masthead.html
+++ b/_includes/masthead.html
@@ -21,9 +21,13 @@
         {% if site.search == true %}
         <button class="search__toggle" type="button">
           <span class="visually-hidden">{{ site.data.ui-text[site.locale].search_label | default: "Toggle search" }}</span>
+          {% if site.search_icon and site.search_icon != 'builtin' %}
+          <i class="{{ site.search_icon }}"></i>
+          {% else %}
           <svg class="icon" width="16" height="16" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 15.99 16">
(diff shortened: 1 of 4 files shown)


iBug:
Looks good at first glance but I feel uneasy. Maybe `default` is a better choice than `builtin`?

rhwood:
I avoided using `default` as the default option because
- `default` is non-descriptive ("the default option is default" does not describe the (current) default behavior)
- I want to provide for the possibility that the default changes from the builtin SVG in the future (specifying `builtin` in `_config.yml` is explicitly ensuring the builtin SVG is used even if the default icon changes to something else later)
- I'm assuming that users who want the default behavior and are unconcerned about the possibility of its changing, will not specify this value and leaving it empty (i.e. by explicitly specifying a value, you are not using the default behavior)

_includes/masthead.html
```diff
@@ -21,9 +21,13 @@
         {% if site.search == true %}
         <button class="search__toggle" type="button">
           <span class="visually-hidden">{{ site.data.ui-text[site.locale].search_label | default: "Toggle search" }}</span>
+          {% if site.search_icon and site.search_icon != 'builtin' %}
+          <i class="{{ site.search_icon }}"></i>
```
iBug, line 25: Can you fix the indentation (add a level) for this line and the 3 lines under (the `<svg>` element)?

iBug (changes requested):

iBug (approved):

rhwood:
Can this please be explicitly denied with an explanation of why the search icon should be different than every other icon or what needs to be done to use a search icon that matches the style of the other icons instead of quietly ignored without comment?

mmistakes:
Adding more feature flags to the `_config.yml` is something I try to avoid. I don't really see a compelling reason here to add another one for something that could easily be done with a local `_includes/masthead.html` override.

rhwood:
It original change wasn’t easily done in that a later upstream change to the masthead.html broke (I forget what, but I had to reimplement this change to incorporate that other change), which is why I proposed this.

Would this be accepted or rejected if it aligned the search icon with the rest of the icon set used without presenting an option to the user?

mmistakes:
Sure, I see value in keeping the icon consistent with the rest of the FA set. So if you want to swap out the inline SVG currently used to match what FA uses that's fine.

PR #2786, opened 2021-01-16
footer: Add ability to hide the RSS Feed item

rlaPHOENiX:

## Summary

In _config.yml atom already exists and is a dictionary, so I could simply add a "show" key, and for backwards compatibility, it will still work as long as `atom.path` isn't empty.

## Context

I want to create a documentation-only page using mm, which won't use jekyll-feed at all, so having a 404'ing RSS feed enforced isn't ideal.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -12,7 +12,9 @@
       {% endfor %}
     {% endif %}
 
-    <li><a href="{% if site.atom_feed.path %}{{ site.atom_feed.path }}{% else %}{{ '/feed.xml' | relative_url }}{% endif %}"><i class="fas fa-fw fa-rss-square" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].feed_label | default: "Feed" }}</a></li>
+    {% if site.atom_feed.show or site.atom_feed.path %}
+      <li><a href="{% if site.atom_feed.path %}{{ site.atom_feed.path }}{% else %}{{ '/feed.xml' | relative_url }}{% endif %}"><i class="fas fa-fw fa-rss-square" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].feed_label | default: "Feed" }}</a></li>
+    {% endif %}
   </ul>
 </div>
 


rlaPHOENiX:
Ideally for the if check, do a check if "show" in atom exists, instead of having to fallback to atom.path, as I see now that atom.path is quite optional and defaults to "feed.xml" meaning it would cause backwards compatibility problems.

SeanKilleen:
In case you or others aren't aware, a quick workaround:

* In your code, create an `_includes` folder if one doesn't exist.
* Copy the contents of <https://raw.githubusercontent.com/mmistakes/minimal-mistakes/master/_includes/footer.html> into a `footer.html` file in that folder.
* Remove the line with the `<li>` that shows the feed icon.

This will override the theme and cause the RSS feed link to disappear.

rlaPHOENiX:
> In case you or others aren't aware, a quick workaround:
> 
>     * In your code, create an `_includes` folder if one doesn't exist.
> 
>     * Copy the contents of https://raw.githubusercontent.com/mmistakes/minimal-mistakes/master/_includes/footer.html into a `footer.html` file in that folder.
> 
>     * Remove the line with the `<li>` that shows the feed icon.
> 
> 
> This will override the theme and cause the RSS feed link to disappear.

Yes but that is in no way ideal.

SeanKilleen:
Sure -- just letting you know there's a relatively straight-forward way to accomplish what you're looking for. Mentioning this because I saw this issue as I came here to grab the footer code and do exactly that for a site I'm working on.

rlaPHOENiX:
> Sure -- just letting you know there's a relatively straight-forward way to accomplish what you're looking for. Mentioning this because I saw this issue as I came here to grab the footer code and do exactly that for a site I'm working on.

Right, just I'd prefer to keep to the official theme layouts and such for easier upgrading without forgetting about any manual overrides and such.

SeanKilleen:
I've added a pull request to this as it makes sense to me and didn't seem too involved to implement. Feel free to add your thoughts there.

SeanKilleen:
....sigh. I am so sorry. I was completely mistaken and I had been reading this as an issue instead of a pull request. Silly mistake, I apologize! I think my PR may be slightly more thorough; if you agree, feel free to close this.

rlaPHOENiX:
> ....sigh. I am so sorry. I was completely mistaken and I had been reading this as an issue instead of a pull request. Silly mistake, I apologize! I think my PR may be slightly more thorough; if you agree, feel free to close this.

Yeah dont fret, yours works out better anyway.

PR #2787, opened 2021-01-16
Add toggle option for RSS feed visibility

SeanKilleen:

## Summary

Adds a `hide` property to the `atom_feed` in `_config.yml`. When `hide` is set to `true` (`false` by default), prevents the header and footer from rendering the RSS feed icons.

## Context

Resolves #2786.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -12,7 +12,9 @@
       {% endfor %}
     {% endif %}
 
+    {% if !site.atom_feed.hide %}
     <li><a href="{% if site.atom_feed.path %}{{ site.atom_feed.path }}{% else %}{{ '/feed.xml' | relative_url }}{% endif %}"><i class="fas fa-fw fa-rss-square" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].feed_label | default: "Feed" }}</a></li>
+    {% endif %}
   </ul>
 </div>
 
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -12,9 +12,9 @@
       {% endfor %}
     {% endif %}
 
-    {% if !site.atom_feed.hide %}
-    <li><a href="{% if site.atom_feed.path %}{{ site.atom_feed.path }}{% else %}{{ '/feed.xml' | relative_url }}{% endif %}"><i class="fas fa-fw fa-rss-square" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].feed_label | default: "Feed" }}</a></li>
-    {% endif %}
+    {% unless site.atom_feed.hide %}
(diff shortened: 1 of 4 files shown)


_includes/head.html
```diff
@@ -31,7 +34,7 @@
 <![endif]-->
 
 {% if site.head_scripts %}
-  {% for script in site.head_scripts %}
-    <script src="{{ script | relative_url }}"></script>
-  {% endfor %}
-{% endif %}
+{% for script in site.head_scripts %}
```
iBug, line 37: I don't quite understand the flattening of indentation here.

_includes/head.html
```diff
@@ -2,7 +2,10 @@
 
 {% include seo.html %}
 
-<link href="{% if site.atom_feed.path %}{{ site.atom_feed.path }}{% else %}{{ '/feed.xml' | relative_url }}{% endif %}" type="application/atom+xml" rel="alternate" title="{{ site.title }} Feed">
+{% if !site.atom_feed.hide %}
```
iBug, line 5: Ruby-ish syntax:

```liquid
{% unless site.atom_feed.hide %}
```

_includes/head.html
```diff
@@ -2,7 +2,10 @@
...
+<link href="{% if site.atom_feed.path %}{{ site.atom_feed.path }}{% else %}{{ '/feed.xml' | relative_url }}{% endif %}"
```
iBug, line 6: Would be better if this (and the next) line is indented by a level for code readability.

_includes/footer.html
```diff
@@ -12,7 +12,9 @@
       {% endfor %}
     {% endif %}
 
+    {% if !site.atom_feed.hide %}
     <li><a href="{% if site.atom_feed.path %}{{ site.atom_feed.path }}{% else %}{{ '/feed.xml' | relative_url }}{% endif %}"><i class="fas fa-fw fa-rss-square" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].feed_label | default: "Feed" }}</a></li>
```
iBug, line 16: This - see my comments on `_includes/head.html`.

iBug (changes requested):

rlaPHOENiX:
:+1: Would love to see this, I agree with all comments posted by iBug

_includes/head.html
SeanKilleen: Good catch; this was a mistake due to the auto-linter I'm using being a little over-zealous. 👍 I had generally saved without formatting but must have missed it once or twice. Will revert those pieces.

_includes/head.html
SeanKilleen: Works for me!

SeanKilleen:
@iBug good feedback. I think I've addressed everything you've raised; let me know if you've got more.

iBug (approved):
It looks all good now. Only unsure whether @mmistakes like the addition of an extra config key, but this one is simple and straightforward so I believe there's not much to worry about.

Issue #2766, opened 2020-12-27
Feature Request: Copy Code


jvanderaa:
I'm looking to add a "copy code" button on top of the liquid tags:

```
{% highlight yaml linenos %}
{% raw %}

```

I've attempted a few guides:

https://experimentingwithcode.com/creating-a-jekyll-blog-with-bootstrap-4-and-sass-part-4/
https://www.aleksandrhovhannisyan.com/blog/dev/how-to-add-a-copy-to-clipboard-button-to-your-jekyll-blog/

Both "out of the box" are not functional when running locally. I was able to get a portion working with copy code if I remove the linenos. Otherwise with linenos setup (which I absolutely love in the more recent versions) it copies the entire table including the line numbers. Not sure if there is an elegant solution there.

I may just need to dig in a bit more on the JS front of things to get it truly working, but thought I'd see if there was interest of just adding it to the base of Minimal Mistakes all together.

Issue #2508, opened 2020-04-28
Add a Copy-to-Clipboard button and support bilibili video


sunete:
## Summary
1. add a Copy-to-Clipboard button to the code block.

2. Make the theme support to instert [bilibili](https://www.bilibili.com/) video.

### add button
I think  Copy-to-Clipboard button is necessary for this theme. A lot of posts contain code blocks. Now, this theme doesn't have a clip-to-clipboard button. When someone needs to copy the code,  they need to select all the code and ctrl+C to copy to clipboard. It is a waste of time and inconvenience. Besides,  some characters will be left. But if you add a button to the code block, it will be convenient.

I have searched for some articles to add a button by myself. I haven't learned programming before, so I can't finish it.

### support post bilibili video
Can you make this theme supports insert [bilibili](https://www.bilibili.com/) video which is a very popular video website in China? When I instert a bilibili video in the post, it couldn't adapt to the width. **This problem may not be what most people need.** So I  tried to solve this problem by myself. Finally I made it. 
This is the code in [_includes/video ](https://github.com/sunete/sunete.github.io/blob/9c5a5d5d740886b3f646f6cd14b3e7ca5762434f/_includes/video#L12)
```js
<div class="responsive-video-container">
  <iframe src="//player.bilibili.com/player.html?bvid={{ video_id }}&page=1&as_wide=1&high_quality=1&danmaku=0" scrolling="no" border="0" frameborder="no" framespacing="0" allowfullscreen="true"> </iframe>
</div>
```

iBug:
Hi, thanks for the suggestion. It seems like you've already figured out about the Bilibili video suggestion, so it'd be great if you could wrap it up and submit a pull request. We'd love to see your contribution! You can check out [the documentation on contributing](https://mmistakes.github.io/minimal-mistakes/docs/contributing/) if you're looking for a how-to.

sunete:
OK，I will creat a pull request later. How about the copy-to-clipboard button? Is it possible to add the button in the next version?

mmistakes:
Depends on the solution for doing a copy and paste button for code blocks. The ones I've seen are clunky and don't let you use standard Markdown `(```)` fenced code blocks, which is a deal breaker for me. Also not exactly sold on adding extra JavaScript to the theme for this.

- https://aleksandrhovhannisyan.github.io/blog/dev/how-to-add-a-copy-to-clipboard-button-to-your-jekyll-blog/

[...]

Issue #2338, opened 2019-12-04
Support copy to clipboard on code blocks


TannerYoung:
I searched the issues but was surprised I didn't find one that matched this. Perhaps my search terms were wrong and if so I apologies. Thank you for the wonderful theme :)

## Summary

Support a copy-to-clipboard button on code blocks.

## Motivation

Improves user experience when interacting with documentation that has bash commands or code which is intended to be copied by the user.

## Drawbacks
Potential bloat, as it appears it will need a jquery style solution: 

https://stackoverflow.com/questions/48078199/jekyll-code-snippet-copy-to-clipboard-button

dieghernan:
See https://github.com/mmistakes/minimal-mistakes/discussions/2795

Issue #2246, opened 2019-08-30
Click on Headers for permalink


hasezoey:
## Summary

Click on a Header to change URL to it & copy to clipboard
![permalink](https://user-images.githubusercontent.com/10911626/63998438-62880d00-cb01-11e9-9305-ffa9cf5e6c24.png) (image from https://jekyllrb.com/docs/)

## Motivation

It would make linking to it faster & more reliable (because of no typos)

## Drawbacks

none?

---

PS: does this feature exists already? because i cannot find it in this theme...

iBug:
This shouldn't be hard to implement. I'll copy some code from another website I maintain and see if it works.

iBug:
Great idea! I just implemented the same feature on my website [in this commit](https://github.com/iBug/iBug-source/commit/6d87a6d54588be5d125aa335f10bf801e71c9593), but I'm not sure how to integrate the commit into Minimal Mistakes. Maybe if someone could provide some guidance on where I should insert the code I'd be able to fire up a PR for this.

luispuerto:
I use [AnchorJS](https://www.bryanbraun.com/anchorjs/) to create the feature. I don't know the advantages or drawbacks against the @iBug solution but seems that AnchorJS has more options and it's easier to customize.

mmistakes:
Only drawback I could see is AnchorJS is way more JavaScript (~350 lines unminified) than @iBug's solution, which is about 10 lines unminified.

marksbren:
How can I turn this off? I do not want these permalinks on my long-form blog.

iBug:
> How can I turn this off? I do not want these permalinks on my long-form blog.

@marksbren You can copy the theme's default [`/assets/css/main.scss` file](https://github.com/mmistakes/minimal-mistakes/blob/master/assets/css/main.scss) into your own website at the exact path, and append the following line to hide those permalink buttons. Note that permalinks will still work, just without the button introduced here.

```css
.header-link {
  display: none;
}
```

PR #2812, opened 2021-02-06
Copy code to clipboard

iBug:
Add a "copy to clipboard" button for each `<pre>` block in main body. Code from [my website repo](https://github.com/iBug/iBug-source/blob/d8dc3a0eca3497d4e6653d24db3ac68992b84f20/assets/js/clipboard.js) with small non-functional adjustments.

Live preview available:

- <https://ibug.github.io/minimal-mistakes/docs/quick-start-guide/> (Regular fenced ` ``` ` code blocks) ([Wayback Machine](https://web.archive.org/web/20210206094228/https://ibug.github.io/minimal-mistakes/docs/quick-start-guide/))
- <https://ibug.github.io/minimal-mistakes/markup-syntax-highlighting/> (Jekyll `{% highlight %}` code blocks) ([Wayback Machine](https://web.archive.org/web/20210206101758/https://ibug.github.io/minimal-mistakes/markup-syntax-highlighting/))

Screenshot:

> image

Caveat: There's only one single color written for the button because I assumed that all skin variants use "dark mode" for blocks of code (`<pre>` blocks).

More testing and appearance tweaking is recommended before merging.

## Contexts

This is a very popular feature and requested multiple times.

#2338 #2508 #2766 #2795

## Some technical notes

My approach does not depend on code blocks having IDs, but instead searches for the containing code block starting from the button. This is not resistant to structural changes, e.g. when an adjacent HTML tag is changed or the button is wrapped in another layer of `<span>` or whatever. (but who does that, really?)

The CSS selector I used is `.page__content pre > code` (and the go back a level to locate the `<pre>` element), which *should* be safe from unexpected HTML structures while covering most regular code blocks.

The actual "copy" action is done via `document.execCommand("copy")`, which should have the best compatibility according to [this answer](https://stackoverflow.com/a/30810322/5958455) (quoted below), yet [it's deprecated](https://stackoverflow.com/posts/comments/113723056) in 2020.

> #### Browser Support
>
> The JavaScript [`document.execCommand('copy')`](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) support has grown, see the links below for browser updates:
>
> - Internet Explorer 10+ (although [this document](https://msdn.microsoft.com/en-us/library/ms537834(v=vs.85).aspx) indicates some support was there from Internet Explorer 5.5+).
> - [Google Chrome 43+ (~April 2015)](https://developers.google.com/web/updates/2015/04/cut-and-copy-commands?hl=en)
> - [Mozilla Firefox 41+ (shipping ~September 2015)](https://developer.mozilla.org/en-US/Firefox/Releases/41#Interfaces.2FAPIs.2FDOM)
> - [Opera 29+ (based on Chromium 42, ~April 2015)](https://dev.opera.com/blog/opera-29/#cut-and-copy-commands)

## Credits

- Design inspired by the same feature from [squidfunk/mkdocs-material](https://github.com/squidfunk/mkdocs-material) (MIT License)
- JavaScript modified from [zenorocha/clipboard.js](https://github.com/zenorocha/clipboard.js/blob/971834388cb13d01bbe915f54ccb440490c7f4b4/src/clipboard-action.js) (MIT License)
- CSS implemented independently by @iBug (author)
- The "Copy" button uses an icon from [Font Awesome Free](https://github.com/FortAwesome/Font-Awesome) (CC BY 4.0 License)


carlesfernandez:
Hi @iBug,

I've been playing with this code and works nice. I have some humble (and surely naive) suggestions:

1.- Sometimes, it makes no sense to copy the code block. For instance, the output of a program in a Terminal. So I would like to be able to shut down the "Copy to clipboard" button in certain code blocks.

For instance, adding something like:

https://github.com/gnss-sdr/geniuss-place/blob/225737d40e73a9012a8d08898e66a1bd0f8d8f7e/assets/js/_main.js#L236-L239

So I can do:
````
```
Nobody wants to copy this text to the clipboard
```
{: class="nocopy"}
````
... but I'm sure there are more elegant ways to do that.


2.- The implementation throws an `Uncaught TypeError: Cannot read property 'querySelectorAll' of null` when the page does not have any code block. This change seems to fix it:

```
var elem = document.querySelectorAll(".page__content pre > code");
  if (elem) {
    elem.forEach(function(element, index, parentList) {
       // Locate the <pre> element
       ...
```


3.- Since `document.execCommand("copy")` is deprecated, maybe it's better to have it as a fallback if `navigator.clipboard` is not available:

```
try {
      textarea.select();
      if (document.queryCommandEnabled('copy') && navigator.clipboard) {
        navigator.clipboard.writeText(textarea.value);
      } else {
        success = document.execCommand("copy");
      }
    } catch (e) {
      success = false;
    }
```

It seems that `navigator.clipboard` is present in recent versions of Chrome, Firefox, Safari and Opera.

**Comment by iBug** [comment]
1. It's a good suggestion, though I went for the CSS approach which should provide more compatibility (IE doesn't support `Element.closest`)

   Preview: <https://ibug.github.io/minimal-mistakes/docs/utility-classes/>

2. It's some other code causing this error, not the part you pointed out. There's nothing to be done to that, however.

   <https://github.com/mmistakes/minimal-mistakes/blob/116e29ad676ea81103a6c238a1f680767d738dab/assets/js/_main.js#L125>

3. Accepted and I refactored the code so when taking the `navigator.clipboard` path the entire "create `<textarea>` element" logic is bypassed, improving performance.

**Comment by iBug** [comment]
@carlesfernandez

1. It's a good suggestion, though I went for the CSS approach which should provide more compatibility (IE doesn't support `Element.closest`)

   Preview: <https://ibug.github.io/minimal-mistakes/docs/utility-classes/>

2. It's some other code causing this error, not the part you pointed out. There's nothing to be done to that, however.

   <https://github.com/mmistakes/minimal-mistakes/blob/116e29ad676ea81103a6c238a1f680767d738dab/assets/js/_main.js#L125>

3. Accepted and I refactored the code so when taking the `navigator.clipboard` path the entire "create `<textarea>` element" logic is bypassed, improving performance.

File: assets/js/_main.js [hunk 1]
```diff
@@ -173,6 +173,12 @@ $(document).ready(function() {
       console.warn(thisButton);
       throw new Error("No code block found for this button.");
     }
+
+    // Skip line numbers if present (i.e. {% highlight lineno %})
+    var realCodeBlock = codeBlock.querySelector("td.code");
```
luispuerto, line 189: I need to change this from `td.code` to `td.rouge-code` to make it work.

luispuerto:
Hey!!! 

Really great!!! I really wanted to add something like this to my site!! Thanks a lot!

luispuerto:
ummm… the feature is neat, but only works once in a page, after that you have to reload to copy again form another codeblock.

luispuerto:
Question… why don't you use [this](https://clipboardjs.com/)?

iBug:
@luispuerto 

> ummm… the feature is neat, but only works once in a page, after that you have to reload to copy again form another codeblock.

It's more likely there's something wrong with your copy of the code. The behavior in the demo links on `ibug.github.io` should be in line with what this PR would bring to the theme.

> Question… why don't you use [this](https://clipboardjs.com/)?

Quoting [@mmistakes here](https://github.com/mmistakes/minimal-mistakes/issues/2246#issuecomment-529744769):

> Only drawback I could see is AnchorJS is way more JavaScript (~350 lines unminified) than @iBug's solution, which is about 10 lines unminified.

This is a similarly simple feature that simpler code would be easier to maintain.

luispuerto:
> It's more likely there's something wrong with your copy of the code. The behavior in the demo links on `ibug.github.io` should be in line with what this PR would bring to the theme.

I swear over ms-dos and unix that even your demo is behaving like that. In safari. On Firefox I can't even make it work. On macOS Big Sur and last versions of all the apps involved. I even disabled content blockers and so just in case. On safari only worked once and if I want to make it work again I need to reload the page. And in Firefox doesn't work at all. 

On Chromium works. 

> Quoting [@mmistakes here](https://github.com/mmistakes/minimal-mistakes/issues/2246#issuecomment-529744769):
> 
> > Only drawback I could see is AnchorJS is way more JavaScript (~350 lines unminified) than @iBug's solution, which is about 10 lines unminified.

I totally understand and see you there. That extension is used for several sites like github, or for example for R documentation sites like [this one](https://ggplot2.tidyverse.org/). 

I much prefer your solution to be honest, if finally works. Specially because I like how you implemented the CSS and it's already a complete solution. 
 
> This is a similarly simple feature that simpler code would be easier to maintain.

I really don't know about that, but wouldn't be easier just to update the extension if it continue to be maintained?

iBug:
@luispuerto Thank you for your report. I've reproduced the error in Firefox and I'm looking into it. I don't have Safari, however.

iBug:
@luispuerto It's a stupid typo from somewhere I couldn't recall. Can you test the demo pages now on Safari? It's working in Firefox for me now. Make sure to clear your browser's cache for the updated JavaScript.

luispuerto:
@iBug now works! 👍🏻 thanks a lot! 🚀

iBug:
@mmistakes Given the discussions here and on #2795, #2508 and #2338, as well as that (at least) two people also ported code from this PR into their own repositories, I think there's enough popularity for this feature to get included. Any ideas?

mmistakes:
I have usability and accessibility concerns with how it's currently implimented.

1. No focus or active states. Hard to tell if you can navigate to the button by keyboard and trigger the copy function.
2. No text label to indicate what the icon even does.
3. No feedback that it's done anything after click/tap... e.g. "Copied!" or something to that affect.
4. Any UI text used should probably be included in the `_data` file so it can be localized and/or changed easily.

Then there's the concern of turning this on by default. I could see users not wanting it at all and now we've made it that they need to add a class to every code block to disable. I'd prefer it be optin globally rather than optout.

This would also prevent loading more JavaScript for a feature not used by someone. Though it might be more of a pain to try and conditionally add those scripts to the page(s) which gives me pause on this PR as well.

iBug:
Addressing one concern quickly:

> they need to add a class to every code block to disable

With the current implementation it's not. There's a rule for `@at-root .no-copy &` so one could just add `.no-copy` to any parent element (preferably `<body>`).

Making this an opt-in feature is similarly easy: Require a "copy-enabled" class for a parent element, then there could be a switch to add that class to `<body>`.

mmistakes:
Still hesitant on this one. I can already see the feature requests asking for a config flag to turn it on/off globally.

iBug:
OK, I'll implement that global toggle flag with two modes: Global enable + individual page or block opt-out / global disable + individual page or block opt-in.

One technical question: Since the elements and all their HTML content are produced from JS, how should I integrate i18n into that? There's no direct way to transfer data from `_data/something.yml` to JS.

mmistakes:
Another global flag is exactly what I’m trying to avoid. 

The point I was trying to make was I’m hesitant on adding this feature. Not because it doesn’t add value, but because it’s the sort of thing I expect users to not be happy with the implementation and want ever more customization.

- I want to turn on code copy for all blocks
- I want to turn it off on code blocks on X post
- I want it on this block but not that block
- I do want to use a Kramdown attribute to enable/disable
- Etc

I’d rather leave it to the user to override the theme if they want this feature vs the theme trying to do something core Jekyll or a plugin should do.

Even better this probably should be an enhancement to Jekyll’s `highlight` tag. Similar to how it allows enabling for line numbering.

iBug:
[Jekyll's `highlight` tag][1] is well beyond the scope of a theme (it's Ruby), but for the other concern,

> expect users to not be happy with the implementation and want ever more customization

Users have always been wanting ever more customization as is evident by the amount of Issues you're receiving, not to mention some of them are fantastical.

I propose this feature for integration into the theme most simply because it's been requested multiple times and has the popularity in the user base. There are ones who have already gone one step sooner and customized on their own. And there certainly are more who want it badly but don't have the skill to push that. 99% of the users will have the attitude that "I'll happily enable it (or stay with it) if it's ready and easily reachable".

And there *will* be users content or discontent [for *every* single feature][2], so that argument pretty much applies to everything. IMHO, the rule of thumb should be "if the majority goes fine with it, then it's fine". I believe it's courteous enough to provide a way to turn it off (and it's easy and documented).

  [1]: https://github.com/jekyll/jekyll/blob/889fe4130bdda65b288c34ea8ebbba4dd0105f54/lib/jekyll/tags/highlight.rb#L56
  [2]: https://xkcd.com/1172/

mmistakes:
Another option would be to go the "helper" `_includes` path that I've done with other opt-in features like galleries, feature rows, etc.

I like the following solution where you drop in an include before a code block you want to add copy/paste support. I think this would address most of my concerns with the current implimentation.

https://www.aleksandrhovhannisyan.com/blog/how-to-add-a-copy-to-clipboard-button-to-your-jekyll-blog/

Instead of Emoji icons I'd leverage Font Awesome, make sure the text strings are done in a way that they can be customized via the UI text data file, and maybe come up with a better include name than code header.

mmistakes:
Can't think of anything that would make a global enable doable.
I'm still against that. My view is the theme shouldn't do it, I only suggested the include as a compromise.

This is the sort of thing I think should be a plugin/enhancement to the `{% highlight %}` tag. For example I like how Gatsby does it with Prism remark plugins that extends the core:

https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-remark-prismjs
https://github.com/DSchau/gatsby-remark-code-titles

I could see the `highlight` tag being enhanced or forked to do something similar:
- highlighting specific lines
- add a code title/filename
- enable copy/paste
- enable line numbering

iBug:
As of the last commit Michael's questions are addresses as such:

> 1. No focus or active states. Hard to tell if you can navigate to the button by keyboard and trigger the copy function.

The theme already has CSS for `button:focus`, though `:active` is missing.

I've tested with both latest Edge and Firefox and can confirm that it works with keyboard navigation.

> 2. No text label to indicate what the icon even does.

Added `<span class="sr-only">Copy code</span>`.

> 3. No feedback that it's done anything after click/tap... e.g. "Copied!" or something to that affect.

There are design and implementation concerns so not easy to address quickly... so maybe later?

Again, [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/getting-started/) is a really good reference design for all these kinds of features. I *may* come up with something sensible in the future.

> 4. Any UI text used should probably be included in the `_data` file so it can be localized and/or changed easily.

The only UI text is the `sr-only` span tag. Since the theme isn't localizing SR text for a few places ([example](https://github.com/mmistakes/minimal-mistakes/blob/4.24.0/assets/js/_main.js#L131)), I'm not going to take the blame right here.

---

Most importantly, I would directly dismiss Michael's primary concern of whether users want this feature, as pretty much [every](https://squidfunk.github.io/mkdocs-material/getting-started/), [single](https://just-the-docs.com/docs/configuration/), [theme]()

Correction: Just The Docs has a site-wide switch, which is [enabled by default](https://github.com/just-the-docs/just-the-docs/blob/a251382b7a68acb4eff37774b31750550809cf1f/assets/js/just-the-docs.js#L572) though. The current approach (using uglify-js to squeeze everything into a single `main.min.js` doesn't allow us to follow that implementation easily.

Now that only point 3 is left completely unaddressed, I'm going to green-light this PR so as not to let this nice feature be blocked for too long.

iBug:
> 3. No feedback that it's done anything after click/tap... e.g. "Copied!" or something to that affect.



This will be shipped along with the next release (4.26.1 or 4.27.0).

I'm no frontend web designer nor engineer, so the implementation is quite primitive. Anyways, better than none.

PR #2815, opened 2021-02-09
Document user custom element hooks

tsmanner:

## Summary

For simple css additions to a site, it would be nice to be able to include bare css files in the user site and have a way to add them to a site or page.  The way to style a page with the lowest barriers to entry is to just write css and include the stylesheets in your `<head>` element.  This PR implements optional the `css` key, for `_config.yml` and page YAML Front Matters, which contains a sequence of paths to css files.  [Site css files](https://github.com/tsmanner/minimal-mistakes/blob/custom-css/_includes/head.html#L19) are included first, followed by [page css files](https://github.com/tsmanner/minimal-mistakes/blob/custom-css/_includes/head.html#L24), in the `<head>` element of each page on the site, via the `head.html` include.

Simple documentation of the configuration values can be found in `configuration.md`:
- [Site CSS](https://github.com/tsmanner/minimal-mistakes/blob/custom-css/docs/_docs/05-configuration.md#site-css)
- [Page CSS](https://github.com/tsmanner/minimal-mistakes/blob/custom-css/docs/_docs/05-configuration.md#page-css)

## Context

This is not related to any issues, just a desire to add some simple, custom styled, elements to a page.  My particular case was to add alternating highlighting and underlines to digits in a numeric string to make it easier to pick out byte and nibble boundaries.  It's useful for me at work to help correlate bit numbers to hex digits in some machine generated log files.


iBug:
Sorry but I can hardly see any utility of this PR.

Global CSS is already easy enough to add and is documented here: <https://mmistakes.github.io/minimal-mistakes/docs/stylesheets/#customizing>

For "page CSS", I can think of two ways of doing so:

1. Use [`_includes/head/custom.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/head/custom.html) to add the exact Liquid code you introduced in this PR, which is documented here: <https://mmistakes.github.io/minimal-mistakes/docs/stylesheets/#font-stacks>
2. Add custom classes to the `<body>` element using `page.classes` in Front Matter, then apply some "site CSS" with a selector scoped for `body.some-class`. This is documented here: <https://mmistakes.github.io/minimal-mistakes/docs/layouts/#layout-based-and-user-defined-classes>

In fact, I've used the second approach for a year or so here: <https://github.com/iBug/iBug-source/blob/9033ff61ea0aea9198c16fd1ac837d50485d253f/_sass/landing.scss#L1>

With that described, I'm not in favor of this PR.

Thank you for your interest in Minimal Mistakes theme. I appreciate how it's submitted in a good faith with good coding and documentation practices.

tsmanner:
That makes sense, I was honestly expecting a response similar to this, but have been having trouble finding the "right" place to inject my css.  My hacked solution was to add the `<link>` tag directly to the top of the page I wanted it in, but that was obviously a bad solution.  I'll give those suggestions a shot, and expect that I'll drop the code change and modify the documentation update to point at those and give a quick inline example.

tsmanner:
@iBug your suggestion worked on the first try, it's too bad it's buried in the documentation for using old font versions.  I added some documentation on the Configuration page with a link to more details on the Stylesheets page.  Thanks for the prompt response, and info!

iBug:
Well, this is not the kind of guidance I expected.

Since we're utilizing `_includes/head/custom.html`, the introduction of `{{ site.css }}` is completely redundant - everything could be just laid out plain there, so why bother moving the actual URLs to another file?

For the page CSS part, I was looking for a ProTip block right under the "user-defined classes" section, mentioning the combined use of custom page classes and scoped site CSS [as I've done][1]. Utilizing `_includes/head/custom.html` for page CSS, on the other hand, isn't as bad as that for site CSS, so I'm neutral toward this.

The change in `docs/_docs/05-configuration.md` is also messy - it focuses on adding CSS via `<link>` tags, introducing more-than-necessary Liquid tags, and more importantly, it's at a wrong place. If you look closely, you'll find that the whole page talks about available settings via `_config.yml`. The insertion of overriding a custom include file is absurd. I'd suggest `docs/_docs/06-overriding-theme-defaults.md` as a better place for this text, but be sure to be "generic" about the usage of `head/custom.html` - it could be used for *anything* in the `<head>` part, not just CSS.


  [1]: https://github.com/iBug/iBug-source/blob/master/_pages/index.md#L5

tsmanner:
That's all fair feedback, I've been writing stuff up the way I'm thinking about _using_ it not the way it should be presented to all users.

> Since we're utilizing _includes/head/custom.html, the introduction of {{ site.css }} is completely redundant - everything could be just laid out plain there, so why bother moving the actual URLs to another file?

The thought behind this was for future additions to my own site, it's easier to find and to add a line to the `_config.yml` than copy/paste `<link>` tags in _includes/head/custom.html.  I agree that it is functionally the same, which is why I included both approaches in the longer example in `_docs/_docs/16-stylesheeds.md`.  That said, the way I presented it there is my preference for putting as much site configuration in `_config.yml` as possible, similarly to the way scripts are handled with the `head_scripts`, `footer_scripts`, and `after_footer_scripts`.

I am curious why CSS is expected to be handled entirely differently than JavaScript in regards to configuration?  I am a naive web developer, in my limited experience, they tend to be handled similarly, as they often work together to render the page.  Regardless of that, I care much more about having some example of page CSS configuration than the site wide stuff, so I would be happy even reducing this to a small segment in `docs/_docs/06-overriding-theme-defaults.md` describing how to use the configuration available to have page-specific CSS (and probably JavaScript) via the custom head/footer files.

I've made a quick change to the branch, removing the section from `docs/_docs/05-configuration.md` and working on the typewriting style in `16-stylesheets.md`.

tsmanner:
I've been thinking about this more, and have been trying to find the right spot to put some example of using `_includes/head/custom.html` to include page-specific CSS.  I'm not coming up with any places that feel quite right.  A one or two sentence ProTip in `docs/_docs/06-overriding-theme-defaults.md` is reasonable, but doesn't seem like the right place to put any sort of example.

I don't see any documentation of the html customization points, only some mentions that they exist.  For example, `02-structure.md` mentions that `_includes/footer` and `_includes/head` are for "custom snippets," and `_includes/analytics-providers` mentions "Google and custom."  There is no mention of _what_ in there is customizable, to determine that I had to start digging through the `.html` source files in the repository to see what gets included and seems to have something like "custom" in the name.

I don't think the right place to talk about how to use the built-in customization points is `docs/_docs/06-overriding-theme-defaults.md`, since they're not overriding anything.  Would it be possible to add a page under the `CUSTOMIZATION` sidebar nav that was something like `docs/_docs/XY-custom-elements.md`?  This page would list the files that are already included by the theme, but that are intended to contain user data, like `_includes/head/custom.html`, and give a brief description of how/where they're included.  Including an example or two I think would be helpful.  This would be especially useful for people, like myself, that are using `minimal-mistakes` via `jekyll-remote-theme`.

Something along these lines:
---

### _includes/head/custom.html
This file is included just before the close of the `<head>` tag in the default layout

### _includes/footer/custom.html
This file is included just after the open of the `<footer>` tag in the default layout

iBug:
Given how files under `docs/_docs` are laid out, it's not easy (or clean) to add a new file into that list.

**There are two other places I found suitable for the mentioning of HTML addition points**, `02-structure.md` and `10-layouts.md`. If you find it easier to discover, I'm fine if you add the same mentioning in both `16-stylesheets.md` and `17-javascript.md` as well (just be sure to link to other pages when necessary, so as to reduce the amount of duplicated content).

**About the reason CSS is handled differently than JavaScript**, I go for [Jekyll's built-in processing for Sass/SCSS](https://jekyllrb.com/docs/assets/#sassscss). And since the theme already supports per-page classes at `<body>` level, it's better not to complicate things when you can do it cleanly.

And last but not least, the current way `head_scripts`, `footer_scripts` etc. are laid out is as redundant as it is. The best way of course, is to drop them altogether and advocate the "HTML insertion points" approach. This is too breaking a change that better fits a major release, like `4.100.0` or 5.0.

tsmanner:
I got back to this this morning and made the following changes:

1. Added a line pointing to `custom.html` beneath the `head` and `footer` paths in `02-structure.md` and mentioned creating the files locally in the note above it.
2. Added a section in `10-layouts.md` that discusses `head` and `footer` includes, along with two examples of adding page-specific CSS and JavaScript using them.

I can continue through the other template includes for custom HTML injection points and add sections there for each of them as well, though I probably won't have very good examples to include.  @iBug let me know what you think

iBug (approved):
Looks good to me. Need @mmistakes to make more reviews.

PR #2831, opened 2021-02-19
Remove all references to official public Staticman API instance.

### LINKED ISSUE #2818: Outdated Staticman Docs
## Motivation

1. Staticman's maintainer recommended users to move away from any public API instance in https://github.com/eduardoboucas/staticman/issues/317#issuecomment-565250060.
2. The current section about adding a GitHub bot as a collaborator is a legacy authentication method.  To stop other users using your own GitHub bot, extra work (not documented) needs to be done.  That's too complicated for normal users.  Switching to GitHub App circumvents this problem because one can make it private (so that others can't use it) and limit the repos to which it's installed.

## Suggestion

In https://github.com/eduardoboucas/staticman/issues/367#issuecomment-660315594, Staticman's maintainer suggested to stop mentioning all references to the current public instance.  We should follow suit as the reliability of the public instance is no longer guaranteed.

In this repo, the public instance is used as a fallback URL.

https://github.com/mmistakes/minimal-mistakes/blob/d6444412c63aea5e47241ef536509fb1bfef4830/_includes/comments.html#L38  
https://github.com/mmistakes/minimal-mistakes/blob/d6444412c63aea5e47241ef536509fb1bfef4830/staticman.yml#L6  
https://github.com/mmistakes/minimal-mistakes/blob/7bb68edf0353a72e82a4c39c6b153eca6721a1eb/docs/_docs/05-configuration.md lines 434, 527

The `default` should be removed, and the associated part of the doc has to be rewritten as a consequence.

As the author of the part for Staticman's native GitLab support in v3 in this repo's docs in #2043, I'm going to propose a PR to keep things up to the current situation.

VincentTam:
This is a bug fix.

## Summary

1. removed any references to the public official API instance "api dot staticman dot net" as Staticman's maintains has recommended users to stay away from it.
    1. replaced `api.staticman.net` in the docs with `{your Staticman API URL}` in the docs
    2. (breaking change) removed the fallback to the public official API instance in the Jekyll-HTML template to encourage existing users to self-host their own instance
2. change `v2` to `v[2|3]` in the docs to encourage users to use the v3 API scheme.  For v3 users, the `/v2/encrypt` in `staticman.yml` might be confusing.
3. labelled the existing GitHub bot authorization mechanism as "legacy" in favor of the new GitHub App way, which avoids GitHub's rate API limit.
4. moved the link to Staticman's quick start guide to the first notice: the linked offiical Staticman setup guide has recently been updated.  It covers both v2 and v3.  It's easier for us (theme maintainer/contributor) to refer (potential/existing) users to the official site in order to avoid overloading our docs page with too technical info about Staticman.

## Context

- resolve #2818 
- docs update subsequent to eduardoboucas/staticman.net#13
- Staticman's maintainer's recommendation: https://github.com/eduardoboucas/staticman/issues/317#issuecomment-565250060

### DIFF
--- a/_includes/comments.html
+++ b/_includes/comments.html
@@ -35,7 +35,7 @@
           <div class="page__comments-form">
             <h4 class="page__comments-title">{{ site.data.ui-text[site.locale].comments_label | default: "Leave a Comment" }}</h4>
             <p class="small">{{ site.data.ui-text[site.locale].comment_form_info | default: "Your email address will not be published. Required fields are marked" }} <span class="required">*</span></p>
-            <form id="new_comment" class="page__comments-form js-form form" method="post" action="{{ site.comments.staticman.endpoint | default: 'https://api.staticman.net/v2/entry/' }}{{ site.repository }}/{{ site.comments.staticman.branch }}/comments">
+            <form id="new_comment" class="page__comments-form js-form form" method="post" action="{{ site.comments.staticman.endpoint }}{{ site.repository }}/{{ site.comments.staticman.branch }}/comments">
               <div class="form__spinner">
                 <i class="fas fa-spinner fa-spin fa-3x fa-fw"></i>
                 <span class="sr-only">{{ site.data.ui-text[site.locale].loading_label | default: "Loading..." }}</span>
--- a/docs/_docs/05-configuration.md
+++ b/docs/_docs/05-configuration.md
@@ -527,7 +527,7 @@ By default comment moderation is enabled in `staticman.yml`. As new comments are
 
 To skip this moderation step simply set `moderation: false`.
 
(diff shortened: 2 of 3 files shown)


VincentTam:
@mmistakes Thanks for your status update, but my PR doesn't address the occurrence of the deprecated instance `api.staticman.net` in the legacy Staticman v1.

https://github.com/mmistakes/minimal-mistakes/blob/f86b5fe49b55a712be12b63cfec2c19134f1f6b4/_includes/comments.html#L112

Removing that would require some refactoring of `_config.yml` like #1845 and that would be a breaking change also, but that would resolve the problem once and for all—if you're OK with this, I would also implement this change.

PR #2839, opened 2021-02-22
feature_row can accept an array with feature rows

lsolesen:

## Summary

This pull request makes the `feature_row` more versatile, so you can add rows without having to specify the rows in the frontmatter.

## Context

There might be many situations where you want to store your feature rows other places than in the frontmatter and be able to query your rows. This makes it possible to still use the standard theme.

### DIFF
--- a/_includes/feature_row
+++ b/_includes/feature_row
@@ -1,5 +1,7 @@
 {% if include.id %}
   {% assign feature_row = page[include.id] %}
+{% elsif include.rows %}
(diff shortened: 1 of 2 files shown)


lsolesen:
I think this is beneficial, as it makes it possible to also use `rows=page.feature_row` or pull your feature rows from other places.

lsolesen:
@mmistakes I think this should be kept open, as it will make it more flexible to use.

PR #2840, opened 2021-02-22
Make it possible to add an array of images

lsolesen:

## Summary

This pull request makes the `gallery` more versatile, so you can add rows without having to specify the rows in the frontmatter.

## Context

There might be many situations where you want to store your gallery images other places than in the frontmatter and be able to query your images. This makes it possible to still use the standard theme.

### DIFF
--- a/_includes/gallery
+++ b/_includes/gallery
@@ -1,5 +1,7 @@
 {% if include.id %}
   {% assign gallery = page[include.id] %}
+{% elsif include.images %}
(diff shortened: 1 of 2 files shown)


lsolesen:
I think this will be beneficial, as it will make it more versatile how to use the gallery, as you can also use `images=page.gallery`

lsolesen:
I think this should be kept open. It makes it much more flexible to use.

PR #2841, opened 2021-02-22
Made it possible for a rel-attribute for feature_row and gallery

lsolesen:

## Summary

When having external links it is useful to be able to specify the rel-attribute. This makes it possible for `feature_row` and `gallery`.


lsolesen:
I think this is beneficial, as when using external links it will be good if you could hint the `rel`-attribute for SEO.

PR #2843, opened 2021-02-24
Multiple Sidebar Nav Lists

tsmanner:

## Summary

This implements a small change to the `nav_bar` include that iterates over elements in the `page.sidebar.nav` variable, creating nav links for each of them.  Existing behavior is unchanged, the new behavior allows the following YAML as well.

```yaml
sidebar:
  nav:
    - main
    - docs
```

## Context

This is not related to any issues.  It is a small change to the include file that allows for more configuration reuse for common navigation elements.

Let me know what you think.  I have one thought already which is that I'm not sure if the `nav_bar` loop should be inside or outside the `<ul class="nav__items">` tag, or possibly even outside the `<nav class="nav__list">` tag with a sidebar.nav.[nav].title key or something like that.  Doing it that way seemed to be introducing a lot more complexity without gaining much usefulness, so I went for the simplest solution here.


iBug:
Have you tested this? I'm afraid it breaks existing setup.

tsmanner:
Huh, I tested it, but with something very simple, I'll take another look today with some more complexity.  I'm okay with closing this in the meantime if you'd like to unclutter the list a bit, I can always reopen it later.

iBug:
Looks OK to me.

The point I missed was that looping over a string in Liquid is like looping over an array with just that string.

tsmanner:
Ah, got it.  It's a pretty handy feature.  I did some more testing on a larger page and it seems to be holding up there.

PR #2844, opened 2021-02-24
Datetime format

luweizheng:

## Summary

When dates are shown on blog posts or pages, a date format will be chosen to format the date string. The default format is "%B %-d, %Y", which will be displayed like “February 24, 2020”.  The default date format is widely used in western world. However, in other cultures like Asian countries, we prefer other formats.

In this PR, I add a date_format variable. If user want to use their own date format, they can add it in `_config.yml`.

## Context

No.


iBug (approved):
This is PERFECT!

Issue #1851, opened 2018-09-26
scrollable sticky toc (enhancement request )


Kulbhushan-Chand:
In reference to **closed issue** on this repo - [Link](https://github.com/mmistakes/minimal-mistakes/issues/1840)

The asked person had made a custom fix.

However If I may request for this feature in the theme,  because having lengthy table of contents overflowed from the page and are not reachable.

@mmistakes  - _**Would you like to introduce it?**_

mmistakes:
Don't know if this could be done gracefully. The right sidebar is a huge hack right now because it supports Markdown auto TOCs and a Liquid helper that creates the TOC.

The placement of the HTML is more flexible with the Liquid helper method, but styling it in a way to not break Kramdown TOCs is challenging.

I prefer the pattern of not sticking the TOC and having it scroll with the content. For documentation heavy sites the left nav sidebar is a better place for adding waypoints.

I'm willing to bet most people ignore the sidebar anyways. There's likely few people who are going to be bouncing around multiple headings on a single page.

Kulbhushan-Chand:
You had me convinced about the added visual clutter because of table of contents. Personally I too give less attention to them. However this may be a dormant feature in the theme. Also the bug in the current `toc `implementation (see the last paragraph of this comment).

My main motive was to get your attention to this closed issue - [**link1**](https://github.com/mmistakes/minimal-mistakes/issues/1840)
where the person had implimented the same on this page - [**link2**](https://wei2624.github.io/MachineLearning/sv_discriminative_model/)

You can see in link 2, he have implemented the scrollable toc. Check this [commit ](https://github.com/Wei2624/Wei2624.github.io/commit/8c8ee0c118c7984f42759b7605da344d859ffbfe)
and the following consecutive commits you can decipher the css changes made by him.

Again I would say that the present system of `toc `in the theme have **problem**  displaying large number of headings in the post. 
**### To see the problem** - You can try to add a lot of headings in the post and in yaml add `toc_sticky: true`. _In this case the table of content will overflow from the page and the last enteries on the list of toc will not be accessible._

mmistakes:
Link2 is super buggy for me. There's all sorts of screen flashes and inconsistencies with how the page scrolls, don't think that's a particularly good solution.

As far as right sidebars that get cutoff. This is one of the reasons I was hesitant to add a sticky behavior. If your body content is shorter than the height of the table of contents, it's going to extend beyond the viewport.

[...]

Issue #1840, opened 2018-09-18
Is there a way to make sticky table of contents scrollable?


Wei2624:
Since I have a long table of contents and I also plan to make it sticky as well, I am wondering if it is possible that I can make right sidebar scrollable?

Thanks

mmistakes:
Not without some custom CSS.

dayne:
I just ran into this issue last week too, wish I'd swung by issue tracker at that time when this ticket was open. 

At the time I was thinking the solution was bringing in a tool like [scrollmagic.io](http://scrollmagic.io/ ) to allow a long right sidebar TOC to be scrollable and sticky.  Sounds like there might be some other approach than a javascript widget.

@Wei2624 did you solve the issue using custom CSS for your right sidebar or did you just fall back to using the left sidebar?

iBug:
This is a really critical feature for those building documentations with this theme. Also easy to implement in the meantime (~10 lines of CSS).

lenhart:
> This is a really critical feature for those building documentations with this theme. Also easy to implement in the meantime (~10 lines of CSS).

@iBug : do you by chance have it implemented and could share the lines of css? It seems to be a reoccurring problem and would thereby getting a step nearer to being solved for good :-)

PR #2874, opened 2021-03-11
Enable toc sidebar scrolling

j3soon:

## Summary

Enable Table of Content sidebar scrolling as mentioned in issue #1851

This can be achieved by simply adding the followings in `_navigation.scss`:

```css
.sidebar__right.sticky > .toc > .toc__menu {
  overflow-y: auto;
  max-height: calc(100vh - 7em);
}
```

When the toc sidebar is not sticky, this rule will not apply. While if the toc sidebar is sticky (`.sticky`), the `toc_menu`'s height will be limited, and the content will be scrollable with a thin-width scrollbar. I use `max-height` instead of `height` to ensure that short TOCs will not be affected by this rule.

Other scrollbar CSS rules is just for making the scrollbar thin in Chrome/Firefox/... If the default wide scrollbar is not preferred, I can simply remove those style rules and keep the 4 lines above.

### Before

image

### After

image

## Context

- #1840 
- #1851

### DIFF
--- a/_sass/minimal-mistakes/_navigation.scss
+++ b/_sass/minimal-mistakes/_navigation.scss
@@ -201,7 +201,7 @@
     }
   }
 
-  img{
+  img {
     -webkit-transition: none;
     transition: none;
   }
--- a/_sass/minimal-mistakes/_navigation.scss
+++ b/_sass/minimal-mistakes/_navigation.scss
@@ -200,7 +200,7 @@
       margin-left: 0;
     }
   }
-  
+
   img{
     -webkit-transition: none;
     transition: none;
@@ -572,18 +572,24 @@
   }
 }
 
-.sidebar__right.sticky > .toc > .toc__menu {
-  overflow-y: auto;
-  scrollbar-width: thin;
-  max-height: calc(100vh - 7em);
-}
-.sidebar__right.sticky > .toc > .toc__menu::-webkit-scrollbar {
-  width: 8px;
-}
-.sidebar__right.sticky > .toc > .toc__menu::-webkit-scrollbar-track {
-  background: #F0F0F0;
-}
-.sidebar__right.sticky > .toc > .toc__menu::-webkit-scrollbar-thumb {
-  background-color: #CDCDCD;
-  border: 1px solid #F0F0F0;
+.sidebar__right {
+  &.sticky {
+    .toc {
+      .toc__menu {
+        overflow-y: auto;
+        scrollbar-width: thin;
+        max-height: calc(100vh - 7em);
+        &::-webkit-scrollbar {
+          width: 8px;
+        }
+        &::-webkit-scrollbar-track {
+          background: #F0F0F0;
+        }
+        &::-webkit-scrollbar-thumb {
+          background-color: #CDCDCD;
+          border: 1px solid #F0F0F0;
+        }
+      }
+    }
+  }
 }
--- a/_sass/minimal-mistakes/_navigation.scss
+++ b/_sass/minimal-mistakes/_navigation.scss
@@ -571,3 +571,19 @@
     padding-left: 3.25rem
   }
 }
+
+.sidebar__right.sticky > .toc > .toc__menu {
+  overflow-y: auto;
+  scrollbar-width: thin;
+  max-height: calc(100vh - 7em);
+}
+.sidebar__right.sticky > .toc > .toc__menu::-webkit-scrollbar {
+  width: 8px;
+}
+.sidebar__right.sticky > .toc > .toc__menu::-webkit-scrollbar-track {
+  background: #F0F0F0;
+}
+.sidebar__right.sticky > .toc > .toc__menu::-webkit-scrollbar-thumb {
+  background-color: #CDCDCD;
+  border: 1px solid #F0F0F0;
+}
--- a/_sass/minimal-mistakes/_navigation.scss
+++ b/_sass/minimal-mistakes/_navigation.scss
@@ -571,25 +571,3 @@
     padding-left: 3.25rem
   }
 }
-
-.sidebar__right {
-  &.sticky {
-    .toc {
-      .toc__menu {
-        overflow-y: auto;
-        scrollbar-width: thin;
-        max-height: calc(100vh - 7em);
-        &::-webkit-scrollbar {
-          width: 8px;
-        }
-        &::-webkit-scrollbar-track {
-          background: #F0F0F0;
-        }
-        &::-webkit-scrollbar-thumb {
-          background-color: #CDCDCD;
-          border: 1px solid #F0F0F0;
-        }
-      }
-    }
-  }
-}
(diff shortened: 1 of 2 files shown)


iBug (changes requested):
This doesn't look like good Sass/SCSS. I recommend nesting and relocating them as needed. For example:

Bad SCSS:

```scss
.example-1 {
  color: #aaa;
}

.example-1 .example-2 {
  font-weight: bold;
}

.example-1 .example-2::before {
  position: absolute;
}
```

Good SCSS:

```scss
.example-1 {
  color: #aaa;

  .example-2 {
    font-weight: bold;

	&::before {
	  position: absolute;
	}
  }
}
```

_sass/minimal-mistakes/_navigation.scss
```diff
@@ -571,3 +571,19 @@
     padding-left: 3.25rem
   }
 }
+
+.sidebar__right.sticky > .toc > .toc__menu {
```
iBug, line 575: Also it's going to be ugly on mobile devices where this whole thing is unneeded. Should probably be wrapped inside 

Perhaps [my working code](https://github.com/iBug/iBug-source/blob/cb512482551e516dfe8049c508b5f6550300c337/_sass/toc.scss) ([preview](https://ibug.io/p/23) and [Wayback Machine](https://web.archive.org/web/20210205163606/https://ibug.io/blog/2021/01/linux-container-explained/)) would be a good reference. But keep in mind that those codes are CC BY-SA 4.0 until I actively contribute them to this repo.

iBug (changes requested):

j3soon:
Thanks for the teaching me the good coding style! I have just refactored my code. I have tested the code on Mobile & PC and on Firefox & Chrome.

iBug:
[`_sidebar.scss`](https://github.com/mmistakes/minimal-mistakes/blob/master/_sass/minimal-mistakes/_sidebar.scss) is definitely a better fit for this change, and that's exactly what I meant by "relocating".

iBug:
Actually this feature is hard to test for lacking a CSS standard of customizing the scroll bar, and that's why I'm hesitant to submit my code as a PR.

j3soon:
The CSS style should work for all major browsers except for Internet Explorer:
- [`scrollbar-width`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width)
- [`::-webkit-scrollbar`](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar)

If such scrollbar customization is not preferred, maybe we can add the TOC scrolling feature with default scrollbar style?

iBug:
@mmistakes What are your comments for this feature? It looks and has been looking nice.

mmistakes:
I'm cool with the PR, but let's remove the custom scrollbar for now and just stick with the default appearance.

mmistakes (changes requested):
Stick with the default scrollbar appearance and remove the custom `::-webkit-scrollbar-*` declarations and I think we're good to merge it in.

iBug:
> I'm cool with the PR, but let's remove the custom scrollbar for now and just stick with the default appearance.

I'm not confident about this. The default appearance looks terrible in this context

> image

j3soon:
I also prefer the custom scrollbar (tested on Chrome/Firefox on PC/Mobile). The default scrollbar looks bad, and without this feature, the overflowing TOC sidebar looks terrible too...

iBug:
This is nice to have. Any reason we can't proceed?

_sass/minimal-mistakes/_sidebar.scss
```diff
@@ -89,6 +89,24 @@
       position: sticky;
       top: 2em;
       float: right;
+
+      .toc {
+        .toc__menu {
+          overflow-y: auto;
+          scrollbar-width: thin;
+          max-height: calc(100vh - 7em);
+          &::-webkit-scrollbar {
```
mmistakes, line 98: Add blank line to remain consistent with spacing

_sass/minimal-mistakes/_sidebar.scss
```diff
@@ -89,6 +89,24 @@
...
+            width: 8px;
+          }
+          &::-webkit-scrollbar-track {
+            background: #F0F0F0;
```
mmistakes, line 102: Is this color needed? I want to avoid a situation where this scrollbar doesn't match the native browser one applied to the entire page.

_sass/minimal-mistakes/_sidebar.scss
```diff
@@ -89,6 +89,24 @@
...
+          }
+          &::-webkit-scrollbar-thumb {
+            background-color: #CDCDCD;
```
mmistakes, line 105: Similar comments to above. Are the colors needed if they're already picking up the default scrollbar colors?

_sass/minimal-mistakes/_navigation.scss
```diff
@@ -200,7 +200,7 @@
       margin-left: 0;
     }
   }
-  
+
   img{
```
mmistakes, line 204: While you're editing this file can you add a space between `img` and `{`?

mmistakes:
The other concern apart from having to maintain custom styles to make the scrollbar match the default one... is the usability of a thin scrollbar on the TOC.

The touch targets are already small on all those TOC links. Throwing in a thin scroll bar seems like we're asking for trouble when used with pointing devices that don't have a scroll wheel/button and you have to actually drag the scrollbar.

mmistakes:
Which browsers are showing the thicker scrollbars BTW? In my limited tests the TOC scrollbar matches the default system one without any `scrollbar-` CSS to make them thinner or change the colors.

**Chrome:**
Screen Shot 2021-07-23 at 10 23 37 AM

**Safari:**
Screen Shot 2021-07-23 at 10 24 33 AM

**Firefox:**
Screen Shot 2021-07-23 at 10 25 52 AM

iBug:
macOS's default scrollbar looks beautiful, but not that of Windows.

mmistakes:
Splitting hairs at this point. I'm not at all bothered by Windows' default scrollbar.
I'd rather not customize one scrollbar and not the other. Let's pull out the custom scrollbar CSS so I can merge this in.

Then if someone wants to take a stab at custom scrollbars that affect the entire theme (not just the TOC) and factor in all the skin color variations... I'll look at that PR and provide feedback.

j3soon:
I've done removing the custom scrollbar styles.

mmistakes (approved):

mendax1234:
@j3soon Seems that using this style I can't let the scroll bar to be round. This is the code I try
```html
.toc {
        .toc__menu {
          overflow-y: auto;
          scrollbar-width: thin;
          max-height: calc(100vh - 7em);
          &::-webkit-scrollbar {
            width: 8px;
          }
          &::-webkit-scrollbar-track {
            background: #F0F0F0;
          }
          &::-webkit-scrollbar-thumb {
            background-color: #CDCDCD;
            border: 1px solid #F0F0F0;
            border-radius: 4px;
          }
        }
      }
```
Thanks for any help!

PR #2951, opened 2021-05-10
test save

AlexFake1:


alexsnoor:
add svelte modal window, need correction main.js

PR #2959, opened 2021-05-16
feat: use FontAwesome for navigation menu

rhwood:

## Summary

Use a FontAwesome icon for the navigation menu displayed when navigation items exceed the available width. Combined with #2774, this theme should be using FontAwesome characters for all built-in iconography.

## Context

This is not related to any open issue.


mmistakes:
Only negative I see with moving to Font Awesome is the icon transition isn't great. There's no smooth animation from the bar icons to the close icon... it's now super janky as there's no way to ease the two icons.

I think in this case I'd rather stick with the pure CSS menu icon instead of using Font Awesome's set.

iBug:
I'm with @mmistakes. While FA is great, the menu button is at a mature state that opting in for FA brings little advantage over the status quo.

rhwood:
I was thinking more about replaceability of the icon(s) without essentially having to override a significant amount of CSS and/or javascript. By only having to override to two variables to change the icon, I can replace the hamburger icon with the kebab (as used in android and other UIs)

I should also note that I only asked https://github.com/mmistakes/minimal-mistakes/discussions/2957 while working on this--until I dug into the CSS while working on this, I hadn't known the X transform was supposed to be there as I never saw it.

mmistakes:
Closing for the reasons cited above.

PR #2962, opened 2021-05-17
Add pinned post to home page, with example

ehcloninger:

## Summary

Adds the ability to 'pin' a post to the top of the front page, such as a welcome message or the T&Cs for the site. Adds a
new boolean front matter value of 'pinned' that triggers layouts/home.html to search for pinned posts. Includes a sample
of a pinned post.

Pinned posts will also show up in recents as they may or may not be on page 1 if paginated.

## Context

Not related to a Github issue. This was a feature that I wanted for my hobby site: https://my72mgb.com

### DIFF
--- a/_layouts/home.html
+++ b/_layouts/home.html
@@ -2,6 +2,23 @@
 layout: archive
 ---
 
+{% if page.url == "/" %}
+  {% assign entries_layout = page.entries_layout | default: 'list' %}
+  {% assign print_header = true %}
+  <div class="entries-{{ entries_layout }}">
+    {% assign posts = site.posts %}
+    {% for post in posts %}
+      {% if post.pinned %}
+        {% if print_header %}
+          <h2 class="archive__subtitle"><i class="fas fa-thumbtack"></i> Pinned Posts</h2>
+          {% assign print_header = false %}
+        {% endif %}
+        {% include archive-single.html type=entries_layout %}
+      {% endif %}
+    {% endfor %}
+  </div>
+ {% endif %}
+  
 {{ content }}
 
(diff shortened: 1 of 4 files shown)


_layouts/home.html
```diff
@@ -2,6 +2,23 @@
 layout: archive
 ---
 
+{% if page.url == "/" %}
```
iBug, line 5: This doesn't look good to me. Perhaps it should be a page-level switch instead of hard-coded "root path"

_layouts/home.html
```diff
@@ -2,6 +2,23 @@
...
+  {% assign entries_layout = page.entries_layout | default: 'list' %}
+  {% assign print_header = true %}
+  <div class="entries-{{ entries_layout }}">
+    {% assign posts = site.posts %}
+    {% for post in posts %}
+      {% if post.pinned %}
```
iBug, line 11: This is redundant and slow. Use [where expression](https://jekyllrb.com/docs/liquid/filters/):

```liquid
{% assign posts = site.posts | where: 'pinned', true %}
```

_layouts/home.html
```diff
@@ -2,6 +2,23 @@
...
+        {% if print_header %}
```
iBug, line 12: This is also unnecessary and can be done in two other ways:

- If the page-level switch is enabled, then a header can be displayed even with no posts (users should handle this on their own).
- If the filtered result (where expression) is non-empty, then show the header.

iBug (changes requested):

_layouts/home.html
ehcloninger: @iBug I'll look at this on my personal site. I'm not a pro at ruby/jekyll, but I am a dev. Thanks

iBug (commented):
This is not understood...

Issue #2964, opened 2021-05-18
_includes/seo.htm is missing post.excerpt from seo_description 


funvill:
## Environment

- Minimal Mistakes version: 4.23.0 
- Git repository URL: https://github.com/funvill/funvill.github.io 
- GitHub Pages hosted (if yes provide URL to site): https://blog.abluestar.com/

## Expected behavior

When linking to a "post" in a system that supports Open Graph such as Slack or Twitter. The site description is used instead of the first paragraph of the "post" as expected. 

The first paragraph is used as an Open Graph description (`og:description`) for pages and collections as expected. 

I expected the default Open Graph description of a post to be the first paragraph of the post instead of the site description. 

## Fix. 

[Line 23 of _includes/seo.htm](https://github.com/mmistakes/minimal-mistakes/blob/e134d13e457a71900f1d5732cb8668205627ba66/_includes/seo.html#L23)  needs to be updated

From 

`{%- assign seo_description = page.description | default: page.excerpt | default: site.description -%}`

to 

`{%- assign seo_description = page.description | default: page.excerpt | default: post.excerpt | default: site.description -%}`

funvill:
as @iBug pointed out in this pull request https://github.com/mmistakes/minimal-mistakes/pull/2965 we can not rely on `post.excerpt` as it's not an "automatic variable in Jekyll".

Upon more research into this issue using the demo https://mmistakes.github.io/minimal-mistakes/ theme, it doesn't appear to be an issue on the demo site. So it must be unique to my installation. I am going to close this issue do more research and open a new issue if needed. 

Thanks for your help @iBug

mmistakes:
Can't reproduce this on the demo site, so maybe there's something specific going on in your repo.

Example post on the demo site: https://mmistakes.github.io/minimal-mistakes/jekyll/welcome-to-jekyll/
And looking at the sourcecode I see the following, which is taking the first paragraph of text and using that as `page.excerpt` by Jekyll.

```html
<meta name="description" content="You’ll find this post in your _posts directory. Go ahead and edit it and re-build the site to see your changes. You can rebuild the site in many different ways, but the most common way is to run jekyll serve, which launches a web server and auto-regenerates your site when a file is updated.">
```

And on a post with a custom excerpt it is always working as expected.

https://mmistakes.github.io/minimal-mistakes/layout/uncategorized/layout-excerpt-defined/

```html
[...]

PR #2965, opened 2021-05-18
Fix issue where post.excerpt was not being used for Open Graph descri…

funvill:
This is a bug fix.

## Summary

This fixes this issue where Open Graph description (og:description) on posts uses the site description instead of the post.excerpt.
See issue https://github.com/mmistakes/minimal-mistakes/issues/2964

### DIFF
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -20,7 +20,7 @@
   {%- assign canonical_url = page.url | replace: "index.html", "" | absolute_url %}
 {% endif %}
 
-{%- assign seo_description = page.description | default: page.excerpt | default: site.description -%}
+{%- assign seo_description = page.description | default: page.excerpt | default: post.excerpt | default: site.description -%}
 {%- if seo_description -%}
   {%- assign seo_description = seo_description | markdownify | strip_html | newline_to_br | strip_newlines | replace: '<br />', ' ' | escape_once | strip -%}
 {%- endif -%}


iBug:
There's already `page.excerpt`. `post` has never been an automatic variable in Jekyll so you may be fixing the wrong thing.

funvill:
@iBug I am new to Jekyll theme updates. 

From this page https://jekyllrb.com/docs/posts/#post-excerpts it sounds like `post.excerpt` is a variable. 

I have tested this fix on my personal website and it seemed to have fixed my issue. 

I updated the issue to be more descriptive

> When linking to a "post" in a system that supports Open Graph such as Slack or Twitter. The site description is used instead of the first paragraph of the "post" as expected. 
> 
> The first paragraph is used as an Open Graph description (`og:description`) for pages and collections as expected. 
> 
> I expected the default Open Graph description of a post to be the first paragraph of the post instead of the site description.

iBug:
> From this page [jekyllrb.com/docs/posts/#post-excerpts](https://jekyllrb.com/docs/posts/#post-excerpts) it sounds like `post.excerpt` is a variable.

No, `post` is a variable because there's `{% for post in site.posts %}` outside. As I said, `post` is not an "automatic variable" that gets set based on context, unlike `page` or `site`.

I don't understand how it's an issue in the first place. Both `page.excerpt` and `page.description` should cover the OG metadata already well.

funvill:
> No, post is a variable because there's {% for post in site.posts %} outside. As I said, post is not an "automatic variable" that gets set based on context, unlike page or site.

@iBug okay it sounds like my fix is not good then. 

I will close this pull requests and update the issue with more details and descriptions. Thanks for your feedback.

PR #2967, opened 2021-05-19
Loads font-awesome asynchronously

ghusse:

## Summary

Loading font-awesome asynchronously allows to display the site faster. This change is advised by google pagespeed insights


iBug:
To be honest, I was thinking that Michael would take this as "too minor".

PR #3014, opened 2021-06-23
Trucated seo_description with recommended length for Google

lsolesen:
It is recommended in most SEO guides only to write 160 characters in the meta description.

When the theme defaults to using the excerpt sometimes the meta descriptions gets really long. Therefore maybe it is a good idea to truncate the meta description.

### DIFF
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -20,9 +20,12 @@
   {%- assign canonical_url = page.url | replace: "index.html", "" | absolute_url %}
 {% endif %}
 
-{%- assign seo_description = page.description | default: page.excerpt | default: site.description -%}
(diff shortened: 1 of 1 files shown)


mmistakes:
I think truncating it is unnecessary. The 160 character limit is an artificial guideline SEOs came up with to target when you're writing your meta descriptions.

Google and other search engines may use all of that text, completely ignore it, rewrite it, use text from the middle of your page instead, or truncate it themself. Passing more text does no harm.

iBug:
That 160-char limit is just a "recommendation", so @mmistakes is right that passing more text does no harm. Better leave this decision for users.

PR #3051, opened 2021-07-09
Remove site.url from first breadcrumb link

### LINKED ISSUE #3050: Breadcrumbs top-level link is not using relative URL but site.url
## Environment

- Minimal Mistakes version: `4.24.0`
- Ruby gem or remote theme version: `4.24.0`
- Jekyll version: `4.2.0`
- Git repository URL: https://github.com/mmistakes/minimal-mistakes
- Operating system: macOS

## Expected behavior
The first link in the breadcrumb, "Home" in the picture below
Screen Shot 2021-07-09 at 23 38 17
should not depend on the `site.url`.

When the site is deployed to `site.url` it works all good, the issue is when e.g. running locally with jekyll-serve, or deploying a preview version at a different URL. In this case, the "Home" link in this breadcrumb is linking to the real page instead of the local/preview url.

With `site.url: domain.com` and running `$ jekyll serve`,  the "Home"
* will point to `http://domain.com/`
* but should point to `http://127.0.0.1/`

The rest of the breadcrumbs works correctly and make relative links already, in this case to `http://127.0.0.1/**/*`. It's just this very top/first/root URL that is wrong, which is visible in the code as well:
https://github.com/mmistakes/minimal-mistakes/blob/3c075fe76b0971f3ccc63c6940b63306d6100fb2/_includes/breadcrumbs.html#L22

This is rather annoying because without noticing I'm at domain.com but still believing I'm clicking around in my test instance. 

Just set `breadcrumbs: true` in `_config.yml`

erikw:
This is a bug fix. 

## Summary
Remove `site.url` from the top-level breadcrumb, so that the URL becomes relative just like the other (child) breadcrumb links.

This means that if a Jekyll site configured as `site.url: domain.com` the top/first/root breadcrumb will now point to `/` instead of `domain.com/`. This is important because the site might be deployed to `127.0.0.1` for local development or at `dev.domain.com` or such. Thus it should simply link to the root relative and not with `site.url`.

## Context

Fixes #3050

### DIFF
--- a/_includes/breadcrumbs.html
+++ b/_includes/breadcrumbs.html
@@ -19,7 +19,8 @@
     {% for crumb in crumbs offset: 1 %}
       {% if forloop.first %}
         <li itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem">
-          <a href="{{ site.baseurl }}/" itemprop="item"><span itemprop="name">{{ site.data.ui-text[site.locale].breadcrumb_home_label | default: "Home" }}</span></a>
+          <a href="{{ '/' | relative_url }}" itemprop="item"><span itemprop="name">{{ site.data.ui-text[site.locale].breadcrumb_home_label | default: "Home" }}</span></a>
+
           <meta itemprop="position" content="{{ i }}" />
         </li>
         <span class="sep">{{ site.data.ui-text[site.locale].breadcrumb_separator | default: "/" }}</span>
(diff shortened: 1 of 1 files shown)


mmistakes (changes requested):
Since you're editing this file would you mind removing `site.baseurl` too and use a `relative_url` filter instead?

For example instead of `{{ site.baseurl }}/` replace with `{{ '/' | relative_url }}`

This is how I've handled root relative URLs that point back to the home page in [other parts of the theme](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/masthead.html#L8) and would help keep things consistent.

erikw:
@mmistakes done. PR is updated!

mmistakes (approved):
Looks great. Approved!

PR #3052, opened 2021-07-10
Add Microformats

dltj:

## Summary

Add microformats to markup

## Context

See https://github.com/mmistakes/minimal-mistakes/discussions/3017

### DIFF
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -1,28 +1,32 @@
 {% assign author = page.author | default: page.authors[0] | default: site.author %}
 {% assign author = site.data.authors[author] | default: author %}
 
-<div itemscope itemtype="https://schema.org/Person">
+<div itemscope itemtype="https://schema.org/Person" class="h-card">
+
+  <a class="u-url hidden" rel="me" href="{{ '/' | relative_url }}">
+    Hidden microformat entry for representative h-card
+  </a>
 
   {% if author.avatar %}
     <div class="author__avatar">
       {% if author.home %}
         <a href="{{ author.home | relative_url }}">
-          <img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image">
+          <img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image" class="u-photo">
         </a>
       {% else %}
-        <img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image">
+        <img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image" class="u-photo">
       {% endif %}
     </div>
   {% endif %}
 
   <div class="author__content">
     {% if author.home %}
-      <a href="{{ author.home | relative_url }}"><h3 class="author__name" itemprop="name">{{ author.name }}</h3></a>
+      <a href="{{ author.home | relative_url }}" class="u-url"><h3 class="author__name" itemprop="name" class="p-name">{{ author.name }}</h3></a>
     {% else %}
-      <h3 class="author__name" itemprop="name">{{ author.name }}</h3>
+      <h3 class="author__name" itemprop="name" class="p-name">{{ author.name }}</h3>
     {% endif %}
     {% if author.bio %}
-      <div class="author__bio" itemprop="description">
+      <div class="author__bio p-note" itemprop="description">
         {{ author.bio | markdownify }}
       </div>
     {% endif %}
@@ -33,7 +37,7 @@
     <ul class="author__urls social-icons">
       {% if author.location %}
         <li itemprop="homeLocation" itemscope itemtype="https://schema.org/Place">
-          <i class="fas fa-fw fa-map-marker-alt" aria-hidden="true"></i> <span itemprop="name">{{ author.location }}</span>
+          <i class="fas fa-fw fa-map-marker-alt" aria-hidden="true"></i> <span itemprop="name" class="p-locality">{{ author.location }}</span>
         </li>
       {% endif %}
 
@@ -55,7 +59,7 @@
 
       {% if author.email %}
         <li>
-          <a href="mailto:{{ author.email }}" rel="me">
+          <a href="mailto:{{ author.email }}" rel="me" class="u-email">
             <meta itemprop="email" content="{{ author.email }}" />
             <i class="fas fa-fw fa-envelope-square" aria-hidden="true"></i><span class="label">{{ site.data.ui-text[site.locale].email_label | default: "Email" }}</span>
           </a>
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -3,28 +3,18 @@
 
 <div itemscope itemtype="https://schema.org/Person" class="h-card">
 
-  <a class="u-url hidden" rel="me" href="{{ '/' | relative_url }}">
-    Hidden microformat entry for representative h-card
-  </a>
-
   {% if author.avatar %}
     <div class="author__avatar">
-      {% if author.home %}
-        <a href="{{ author.home | relative_url }}">
-          <img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image" class="u-photo">
-        </a>
-      {% else %}
+      <a href="{{ author.home | default: '/' | absolute_url }}">
         <img src="{{ author.avatar | relative_url }}" alt="{{ author.name }}" itemprop="image" class="u-photo">
-      {% endif %}
+      </a>
     </div>
   {% endif %}
 
   <div class="author__content">
-    {% if author.home %}
-      <a href="{{ author.home | relative_url }}" class="u-url"><h3 class="author__name" itemprop="name" class="p-name">{{ author.name }}</h3></a>
-    {% else %}
-      <h3 class="author__name" itemprop="name" class="p-name">{{ author.name }}</h3>
-    {% endif %}
+    <h3 class="author__name p-name" itemprop="name">
+      <a class="u-url" rel="me" href="{{ author.home | default: '/' | absolute_url }}" itemprop="url">{{ author.name }}</a>
+    </h3>
     {% if author.bio %}
       <div class="author__bio p-note" itemprop="description">
         {{ author.bio | markdownify }}
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -40,14 +40,14 @@
       {% if author.links %}
         {% for link in author.links %}
           {% if link.label and link.url %}
-            <li><a href="{{ link.url }}" rel="nofollow noopener noreferrer"><i class="{{ link.icon | default: 'fas fa-link' }}" aria-hidden="true"></i><span class="label">{{ link.label }}</span></a></li>
+            <li><a href="{{ link.url }}" rel="nofollow noopener noreferrer me"><i class="{{ link.icon | default: 'fas fa-link' }}" aria-hidden="true"></i><span class="label">{{ link.label }}</span></a></li>
           {% endif %}
         {% endfor %}
       {% endif %}
 
       {% if author.uri %}
         <li>
-          <a href="{{ author.uri }}" itemprop="url">
+          <a href="{{ author.uri }}" itemprop="url" rel="me">
             <i class="fas fa-fw fa-link" aria-hidden="true"></i><span class="label">{{ site.data.ui-text[site.locale].website_label | default: "Website" }}</span>
           </a>
         </li>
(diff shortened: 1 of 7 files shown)


_layouts/single.html
```diff
@@ -17,21 +17,30 @@
...
   {% include sidebar.html %}
 
-  <article class="page" itemscope itemtype="https://schema.org/CreativeWork">
+  <article class="page h-entry" itemscope itemtype="https://schema.org/CreativeWork">
     {% if page.title %}<meta itemprop="headline" content="{{ page.title | markdownify | strip_html | strip_newlines | escape_once }}">{% endif %}
     {% if page.excerpt %}<meta itemprop="description" content="{{ page.excerpt | markdownify | strip_html | strip_newlines | escape_once }}">{% endif %}
     {% if page.date %}<meta itemprop="datePublished" content="{{ page.date | date_to_xmlschema }}">{% endif %}
     {% if page.last_modified_at %}<meta itemprop="dateModified" content="{{ page.last_modified_at | date_to_xmlschema }}">{% endif %}
 
+  <ul class="hidden">
```
mmistakes, line 26: Is there a better way of applying microformats for this data? This is the sort of extraneous markup I wanted to avoid, especially since it's duplicating the published/modified datetimes from above along with the authorship that is already being applied the author sidebar.

I get all sorts of "code smells" from it not being DRY and having to hide it all via CSS.

_layouts/single.html
```diff
@@ -17,21 +17,30 @@
...
+    <li>These list items are microformat entries and are hidden from view.</li>
+    <li class="u-url">{{site.url}}{{page.url}}</li>
+    {% if site.author.name %}<li><a class="p-author h-card" href="{{ site.url }}"><img class="photo" src="{{ author.avatar | absolute_url }}">{{ site.author.name | markdownify | strip_html | strip_newlines | escape_once }}</a></li>{% endif %}
```
mmistakes, line 29: Careful with your author variables. You start with `site.author` which will cause problems if someone has assigned an author to the post where it should [override the "site" author](https://mmistakes.github.io/minimal-mistakes/layout-author-override/). 

You're also mixing in a `{{ author }}` variable to do the avatar. They probably should all be `{{ author }}`. Take a look at how the [`assign` works in the author sidebar](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/author-profile.html#L1-L2) to see how I did it there. 

A conditional is needed for `author.avatar` since we can't guarantee someone set an image and wouldn't want an empty ``.

Either way I question if the author microformat markup is even needed here as you've added it to the sidebar already. Will it fail without it? Just trying to get a feel for how DRY this can be since I'm not really thrilled with all the duplicate markup.

mmistakes (changes requested):
Overally happy with this PR. I'm not thrilled with adding duplicate markup in `_layouts/single.html` so if any of that can be removed or DRYed up I'd feel better about it.

_layouts/single.html
dltj: Yes—the repeated variables bother me, too.   The driving cause for that is how microformats need the data to be the contents of the enclosing tag and Schema.org puts the data in attributes.  Specifically:
> microformats do not use `link` or `meta` elements within the content of the page and in some cases require particular elements to be used to encode information, such as using `abbr` to support the datetime-design-pattern as illustrated by the `dtstart` property in the example above.  <cite>("<a href="https://www.w3.org/wiki/Mixing_HTML_Data_Formats#Mixing_Syntaxes">Mixing Syntaxes</a>" heading at W3C's "Mixing HTML Data Formats" page)<cite>

I used existing tags where possible (the `p-name` class for the <h1> article title and the `e-content` class in the article <section> tag).  The contents of the hidden <ul>, though, were the values from the Schema.org <meta> tags.

For a while, I was putting microformat markup in the `_includes/page__meta.html`, but separating the article microformat data across two template files felt like a worse compromise as compared to keeping all of the HTML data formats in close proximity in one template file.

_layouts/single.html
dltj: Oh, wow... an excellent point about the mixing of author variables.  I will fix that with an update to the branch.

Regarding the author microdata markup in two places.  The markup in the sidebar is describing an [h-card](https://microformats.org/wiki/h-card).  In the microformats world, it seems like the site homepage should contain this markup to describe the controlling entity of the site.  The author markup in the "_single.html" file covers the stated author of the article on the page for an [h-entry](http://microformats.org/wiki/h-entry).  (And that is where I'm definitely misusing `{{ site.author }}`.

I'm not familiar with the mechanics of multi-author Jekyll blogs, but from what I can gather the `site` variables are appropriate for the sidebar (the "h-card" microformat covering the entire site) while the `author` variables are appropriate for the article itself (the "h-entry" microformat).

dltj:
Please see the updated commit to the PR.  I'm using the assigned-`author` structure now.  I found that it does not have an author "URL" defined in it, so I removed that from the markup rather than making the assumption that the blog root is the author's URL.  The Microformats crowd would probably strongly prefer that to make explicit the connection between the article author and its "h-card" representation, but it is technically optional.  I could also add an author URL to the sample `_config.yml`, but that seems like overkill for this specialized purpose.

_layouts/single.html
mmistakes: With all of the "hidden" list element items, which are actually required? Can we move forward with the other data you marked up that isn't hidden from view?

I'm still iffy on the duplicating all the author markup as it's in the sidebar with microformat markup now.

Can the published/modified dates be marked up in [`_includes/page_date.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/page__date.html) instead? Neither of those are hidden and you wouldn't have to duplicate the markup then.

_layouts/single.html
```diff
@@ -17,21 +17,26 @@
...
+  <article class="page h-entry" itemscope itemtype="https://schema.org/CreativeWork">
     {% if page.title %}<meta itemprop="headline" content="{{ page.title | markdownify | strip_html | strip_newlines | escape_once }}">{% endif %}
     {% if page.excerpt %}<meta itemprop="description" content="{{ page.excerpt | markdownify | strip_html | strip_newlines | escape_once }}">{% endif %}
     {% if page.date %}<meta itemprop="datePublished" content="{{ page.date | date_to_xmlschema }}">{% endif %}
     {% if page.last_modified_at %}<meta itemprop="dateModified" content="{{ page.last_modified_at | date_to_xmlschema }}">{% endif %}
 
+  <ul class="hidden">
+    <li>These list items are microformat entries and are hidden from view.</li>
+    <li class="u-url">{{ page.url | absolute_url }}</li>
+  </ul>
```
dltj, line 29: Hmm—the specification is fuzzy ("[No properties are explicitly required](https://microformats.org/wiki/h-entry#what_is_the_bare_minimum_list_of_required_properties_on_an_h-entry)") and then goes on to say an "h-entry should have the following properties at a minimum for consumers:"

* `url`
* `dt-published` (timestamp)

In this latest commit, I moved `dt-published` to `page__date.html` as you suggested and removed `dt-updated` (the if..else markup in `page__date.html` doesn't distinguish between published and updated when it is rendered to the page).  I also removed `p-author` and `p-summary` as they aren't in the required subset.  

That leaves `u-url`.  Interestingly, there is a [`{{ page.link}}` in `_layouts/single.html`](https://github.com/mmistakes/minimal-mistakes/blob/master/_layouts/single.html#L44), but I can't find a condition where `page.link` would have a value (unless it is expected to come from the page front matter).  Unfortunately, `u-url` is a required data element.  Thoughts on how to handle this?

> I'm still iffy on the duplicating all the author markup as it's in the sidebar with microformat markup now.

The duplication is necessary so the author information occurs within the enclosing `h-entry` markup, and the sidebar is outside that markup.  But since it isn't required I just removed it.

mmistakes: > The duplication is necessary so the author information occurs within the enclosing h-entry markup, and the sidebar is outside that markup. But since it isn't required I just removed it.

Now I’m remembering why I wasn’t keen on adding microformats years ago 🤣. It would be so much better if it leveraged other standards especially all this data is already in the `head` or other schema used in the SEO include.

The `page.link` variable stuff is for “link” posts so you can’t rely on it. https://mmistakes.github.io/minimal-mistakes/post%20formats/post-link/

Maybe the best path forward is add the bare minimum of microformats that doesn’t duplicate markup. Then let me noodle on future support.

The theme is showing it’s age and a lot of other features being requested could benefit from me doing a complete overhaul. Microformats could be lumped in with that…

dltj: Yes, I am disappointed that the IndieWeb folks aren't using something more structured like Schema.org, but I don't have any leverage to steer that community...and it seems like they have quite the organic mix of tool providers that are using microformats.

Are you okay with merging the pull request as it is now?  The minimal parts of microformats for Card and Entry have been added to the existing theme markup with the exception of the hidden &lt;ul&gt; list with the one list item for the URL to the post.  I could clean up the wording of the list item explaining the hidden list—say, "This list item holds the microformat entry for the link to this post, and it is hidden from view"—and make a final commit to this pull request, if you prefer.

mmistakes: The hidden content still gives me a bad feeling.
In fact was looking at the Microformats wiki and they even discourage it.

> If some content doesn't fit into your visible design, then it's not worth microformatting it.

https://microformats.org/wiki/faq#Hidden_Content

dltj: You're right: the Microformats wiki does discourage the use of hidden content.  That said, I think having something in the markup that points to itself seems to be a pretty critical idea—especially if some machine process is going to extract the data out of the page to make use of somewhere else.

I went poking through the Schema-org markup for [`CreativeWork`](https://schema.org/CreativeWork), and there is a property from the inherited `Thing` for URL—and that property isn't used in the mmistakes markup at the moment.  Perhaps a permalink can be added to the footer of the article, and that can be marked up with both `itemprop="headline"` (for Schema-org) and `class="u-url"` (for Microformats).

image

```html
<p>
  <strong><i class="fas fa-fw fa-bullseye" aria-hidden="true"></i> Permalink:</strong> <span itemprop="url" class="u-url">https://dltj.org/article/dltj-with-webmention/</span>
</p>
```

If you like, I could wrap it in a Jekyll {% if %} block that checks for a feature flag that is set to "false" by default so it doesn't have a visible impact on existing sites.

mmistakes: What if instead of adding it to the footer metas... and adding yet another config toggle, you include the permalink in the `h1` with the post title? I don't think it harms anything doing that.

```liquid
<h1 id="page-title" class="page__title" itemprop="headline">
  <a href="{{ page.url | absolute_url }}" class="u-url">{{ page.title | markdownify | remove: "<p>" | remove: "</p>" }}</a>
</h1>
```

About the only change would be a visual one as the title is now styled like a link... which I don't see any issue with.

Screen Shot 2021-07-22 at 3 36 39 PM

dltj:
Yes, this looks good.  I didn't like the styling of the &lt;h1&gt; tag as a link, so I overrode that in the SCSS to make it look as it did before.

_includes/author-profile.html
```diff
@@ -1,28 +1,32 @@
 {% assign author = page.author | default: page.authors[0] | default: site.author %}
 {% assign author = site.data.authors[author] | default: author %}
 
-<div itemscope itemtype="https://schema.org/Person">
+<div itemscope itemtype="https://schema.org/Person" class="h-card">
+
+  <a class="u-url hidden" rel="me" href="{{ '/' | relative_url }}">
```
mmistakes, line 6: Can we remove this hidden markup since `u-url` is being applied to the `author.home` link.
If `u-url` is required and the thought was to add the hidden markup because there is no guarantee someone adds `author.home` to their config.

Perhaps we can do a similar thing to the post title. In this case make the author name a link... use `author.home` first fallback to the site's root URL if the home link hasn't been configured.

```liquid
<h3 class="author__name" itemprop="name" class="p-name">{{ author.name }}</h3>
  <a class="u-url" rel="me" href="{{ author.home | default: '/' | relative_url }}">{{ author.name }}</a>
</h3>
```

_sass/minimal-mistakes/_page.scss
```diff
@@ -72,6 +72,11 @@ body {
   margin-top: 0;
   line-height: 1;
 
+  a {
+    color: $dark-gray;
```
mmistakes, line 76: Replace with `$text-color` variable so each of the skins can assign the correct text color to match. They're not all dark gray.

mmistakes (changes requested):
Thanks for the work on this PR. We're really close. Just the small Sass variable change and then replacing out the hidden author URL markup and this will be good to merge.

dltj:
Ah, yes—moving that hidden markup into the `author__name` &lt;h3&gt; is a good pattern.  Speaking of good patterns, I also applied the same `{{ author.home | default: '/' | absolute_url }}` construct to the `author__avatar` block immediately above this code to remove the `{% if author.home %}` conditional.

`$text-color` fixed too...that was a midnight mistake.

mmistakes (approved):
All systems are go!

dltj:
Many thanks.  I appreciate working with you on this.

PR #3056, opened 2021-07-11
Feature: Add target attribute for navigation link

xepozz:
This PR adds `target` attribute for links in the navigation menu.

P.S. I think may be add `attributes` property for more flexible configuration these link?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 *~
 *.sw[p_]
 
+# IntelliJ IDEA
+*.idea
+
 # Sublime Text
 *.sublime-project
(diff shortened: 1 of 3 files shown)


xjlin0:
in the meantime, I'd use `[link name](https://www.example.com/){:target="_blank"}`

iBug:
This one is sufficiently small for inclusion, but when the next attribute comes up (e.g. if someone asks for `rel="whatever"`, I'll be eager to deprecate everything except `url` and `title` (which is in fact link text and thus essential) and move all of them to an `attributes` property.

PR #3057, opened 2021-07-11
Feature: Profile telegram link

xepozz:
I've added shortcut for [Telegram](https://telegram.org/)

### DIFF
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -248,7 +248,7 @@
 
       {% if author.telegram %}
         <li>
-          <a href="https://t.me/{{ author.vine }}" itemprop="sameAs" rel="nofollow noopener noreferrer">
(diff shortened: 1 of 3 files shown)


iBug:
I'm uncertain about the utility of this patch. You can already add custom links by adding to `author.links`. Specifying directly `author.something` is what Michael already wanting to deprecate, so I'm skeptical whether this gets favored.

Besides, Telegram's logo is designed to stand out with usual text color. Adding a custom blue-ish color is only slightly better than no, but may be harmful on some skins (color palettes). For example I very much doubt it'll look good on the plum skin.

mmistakes:
@iBug summed it up well. Thanks for the PR but this method of adding author links is deprecated in favor of the more flexible author links array method described in the docs.

https://mmistakes.github.io/minimal-mistakes/docs/configuration/#site-author

PR #3096, opened 2021-07-31
Make it possible to enable breadcrumbs per page

lsolesen:

## Summary

Makes it possible to enable breadcrumbs on a per page basis. That will make it possible to disable breadcrumbs for an entire colletion alos by only editing the _config.yml.


lsolesen:
Changed logic so it is possible to enable instead of disable.

PR #3113, opened 2021-08-08
Do not markdownify title - Fixes #3094

### LINKED ISSUE #3094: Title should not be markdownified
## Environment

- Minimal Mistakes version:
- Ruby gem or remote theme version:
- Jekyll version:
- Git repository URL:
- GitHub Pages hosted (if yes provide URL to site):
- Operating system:

I cannot use a pipe | in the title, as it will be changed by the markdownify. It seems unneccesary to markdownify the title, but there is probably some reason for this?

```
{%- if seo_title -%}
  {%- assign seo_title = seo_title | markdownify | strip_html | strip_newlines | escape_once -%}
{%- endif -%}
```

## Expected behavior

I think it should be pssible to use a pipe in the title.

Just try to add a pipe to the title.

lsolesen:
Title should not be markdownified. Pull request to fix gh-3094.


_includes/seo.html
```diff
@@ -4,14 +4,14 @@
 {%- endif -%}
 {%- assign seo_url = seo_url | default: site.github.url -%}
 
-{% assign title_separator = site.title_separator | default: '-' | replace: '|', '&#124;' %}
+{% assign title_separator = site.title_separator | default: '-' %}
```
iBug, line 7: I'd rather keep this line. Liquid `escape_once` doesn't cover the vertical bar. （[Reference](https://www.rubydoc.info/gems/liquid/4.0.0/Liquid/StandardFilters#HTML_ESCAPE_ONCE_REGEXP-constant))

iBug (changes requested):

iBug:
You can add `Fixes #3094` to your PR description so it's automatically closed when this is merged. See <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword>

lsolesen:
But why should the vertical bar | be escaped in the seo_title?

If it needs replacing then the replace should probably take place on seo_title together with escape_once instead, as the user might use it as part of the title also.

iBug:
Well you're right. I couldn't find any necessity to escape the vertical bar. Maybe it was related to Markdown parsing where it would be interpreted as table. Now that you've removed `markdownify`, this escape shouldn't be needed anymore.

iBug (approved):

mmistakes:
The `markdownify` filter does serve a purpose as it normalizes titles that may use Markdown to apply emphasis or other "light" formatting.

For example check this test post on master. https://mmistakes.github.io/minimal-mistakes/markdown/markup-title-with-markup/

Screen Shot 2021-08-08 at 1 54 14 PM

The title has a word italized and bolded, and the `<title>` element is readable without Markdown characters appearing.
Screen Shot 2021-08-08 at 1 51 59 PM

This PR breaks that. Now the `<title>` element has all the Markdown syntax there. Markdown titles are probably an edge case, but the theme has had support for them for a long time and this breaks that.

Screen Shot 2021-08-08 at 1 53 32 PM

mmistakes:
A better fix might be to apply the Markdown and replace transforms on `page.title` before the separator and `site.title` are appended. Then it won't do unintended things on the `|` or any other characters that serve a purpose in Markdown.

_includes/seo.html
```diff
@@ -4,14 +4,14 @@
...
 
 {%- if page.title -%}
   {%- assign seo_title = page.title | append: " " | append: title_separator | append: " " | append: site.title -%}
 {%- endif -%}
 
 {%- if seo_title -%}
-  {%- assign seo_title = seo_title | markdownify | strip_html | strip_newlines | escape_once -%}
```
mmistakes, line 14: Try applying these filters on `page.title` a few lines above before appending the spaces, `title_separator` and `site.title`.

mmistakes (changes requested):

lsolesen:
Thanks for the comments @iBug and @mmistakes. Here is a new stab on solving the issue.

_includes/seo.html
```diff
@@ -4,10 +4,15 @@
 {%- endif -%}
 {%- assign seo_url = seo_url | default: site.github.url -%}
 
-{% assign title_separator = site.title_separator | default: '-' | replace: '|', '&#124;' %}
+{% assign title_separator = site.title_separator | default: '-' %}
 
 {%- if page.title -%}
-  {%- assign seo_title = page.title | append: " " | append: title_separator | append: " " | append: site.title -%}
```
iBug, line 10: I think the problem lies here: It assigns `page_title` (with fallback values) only when `page.title` is set, seems like circular logic to me.

iBug (changes requested):

_layouts/single.html
```diff
@@ -18,7 +18,7 @@
   {% include sidebar.html %}
 
   <article class="page h-entry" itemscope itemtype="https://schema.org/CreativeWork">
-    {% if page.title %}<meta itemprop="headline" content="{{ page.title | markdownify | strip_html | strip_newlines | escape_once }}">{% endif %}
+    {% if page.title %}<meta itemprop="headline" content="{{ page_title }}">{% endif %}
```
iBug, line 21: I'm afraid `page_title` as calculated in `seo.html` won't propagate to here.

PR #3114, opened 2021-08-08
Enable magnific popup on <a> tags only when it has <img>

### LINKED ISSUE #3111: Magnific popup gets in the way of Wikimedia links
This is a follow-up to #1975, which was closed by its author without a real solution.

## Environment

- Minimal Mistakes version: 4.24.0
- Ruby gem or remote theme version: 4.24.0
- Jekyll version: 4.2.0
- Git repository URL: 
  - `main` branch with markdown source: https://github.com/p3palazzo/tau0005/
  - `gh-pages` branch with generated website: https://github.com/p3palazzo/tau0005/tree/gh-pages
- GitHub Pages hosted (if yes provide URL to site): YES  https://palazzo.arq.br/tau0005/ (redirects from https://p3palazzo.github.io/tau0005/)
- Operating system:
  - Built by Github actions on `ubuntu:latest` using:
  -  [this Makefile](https://github.com/p3palazzo/tau0005/blob/c3eac4f56c8cad9d82c0d8c5d7b923b0a9f9a144/Makefile#L25) and
  -  [this workflow](https://github.com/p3palazzo/tau0005/blob/c3eac4f56c8cad9d82c0d8c5d7b923b0a9f9a144/.github/workflows/build.yaml#L16)
  - Uses `jekyll-pandoc` to process markdown instead of kramdown.

## Expected behavior

Links to Wikimedia commons pages should open as any regular link does. Instead, because the URL ends with `'.jpg'` or `'.png'`, Minimal Mistakes tries to load them as a Magnific Popup lightbox.

1. Add a link to a Wikimedia Commons page such as: `[Harvey Barrison, 2012](https://commons.wikimedia.org/wiki/Image:Balloon_Safari_2012_06_01_3128_(7522677448).jpg)` (from line 72 of [this file in the above-mentioned repository](https://github.com/p3palazzo/tau0005/blob/main/_aula/processos.md#L72))
2. When generating the website, everyting looks fine: `<a href="https://commons.wikimedia.org/wiki/Image:Balloon_Safari_2012_06_01_3128_(7522677448).jpg">Harvey Barrison, 2012</a>` (line 285 on the corresponding [generated HTML file](https://github.com/p3palazzo/tau0005/blob/cff5edfbafa7c35f8742cd7b1d9cf623fc6aa9df/aula/processos/index.html#L285))
2. When loading the web page, JavaScript adds the popup class: `<a href="https://commons.wikimedia.org/wiki/Image:Balloon_Safari_2012_06_01_3128_(7522677448).jpg" class="image-popup">Harvey Barrison, 2012</a>` (can be seen [here](https://palazzo.arq.br/tau0005/aula/processos/#fig:serengeti))
3. However, the link does not point to an image but to a web page, which causes the lightbox to fail with `Image#1 could not be loaded`

## Other

As pointed out by @mmistakes [here](https://github.com/mmistakes/minimal-mistakes/issues/1975#issuecomment-441448682), the lightbox can be disabled by editing [l. 85-87 of the `assets/js/_main.js` file](https://github.com/mmistakes/minimal-mistakes/blob/084ddf8d8e5435b850bdd3b05bd377f02621ab54/assets/js/_main.js#L85). However, having to checkout and the rebuild the whole theme Javascript seems like overkill.

I suppose a regex similar to the one that activates the popup could be used to detect wikimedia URLs and disable the lightbox on them, but it would introduce other problems, e.g. what if someone actually points a link to the *image* itself that is hosted on Wikimedia?

iBug:
This is a bug fix.

## Summary

`<a>` tags without an `` child are generally desired *not* to be magnific popups. Uses [jQuery `.has()` function](https://api.jquery.com/has/) to filter those tags that only has a *direct* child `` element.

## Context

Resolves #3111


assets/js/_main.js
```diff
@@ -84,7 +84,7 @@ $(document).ready(function() {
   // add lightbox class to all image links
   $(
     "a[href$='.jpg'],a[href$='.jpeg'],a[href$='.JPG'],a[href$='.png'],a[href$='.gif'],a[href$='.webp']"
-  ).addClass("image-popup");
+  ).has("> img").addClass("image-popup");;
```
iBug, line 87: @mmistakes I accidentally left an extra semicolon here. Would you mind helping remove it? I'm asking because I wouldn't bother with another PR for this thing too trivial (and it really changes nothing functional).

No rebuild is needed as uglify.js is smart enough to ignore these semicolons and turn them into commas.

PR #3119, opened 2021-08-16
Feature/figure popup

JohnScottUK:

## Summary

Added a popup=true argument to figure helper to allow image to be opened in light box (like gallery images)

## Context


lsolesen:
I think this is helpful and a good idea to be able to let users open the image in a lightbox to see a bigger image if needed.

PR #3123, opened 2021-08-19
Feature/related by tags and categories

JohnScottUK:

## Summary

Add ability to include related pages by tags and/or categories. With the related_threshold > 0 pages with the matching number of tags and/or categories will be included instead of site.related_posts.

## Context

### DIFF
--- a/_layouts/single.html
+++ b/_layouts/single.html
@@ -64,12 +64,46 @@ layout: default
     {% endif %}
   </article>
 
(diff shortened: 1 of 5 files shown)


mmistakes:
Closing as this is a Duplicate of #978, a PR I submitted way back in 2017. It hurts build times pretty bad if you have a big site as looping over `site.posts` to compare against tags or categories is costly in Liquid-land.

Using Jekyll related post plugins instead perform much better in my experience than any amount of Liquid spaghetti conditionals.

PR #3127, opened 2021-08-22
Feature/redirect layout

JohnScottUK:

## Summary

Added in redirect layout when Jekyll sites do not have access to rewrite rules on proxy servers/redirects.  Useful when reorganising a site that has incoming links you want to preserve.

## Context

### DIFF
--- a/_layouts/redirect.html
+++ b/_layouts/redirect.html
@@ -1,6 +1,6 @@
 ---
 # Jekyll layout that redirects to another page.
-# Version v1.0.0
+# Version v1.0.1
 # https://www.jsware.io/jekyll-redirect-html/
 # © 2021 John Scott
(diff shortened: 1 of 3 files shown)


mmistakes:
Instead of adding a custom layout I'd much rather leverage Jekyll core's redirect plugin. It's more standardized, doesn't involve any layouts, and has more features.

https://github.com/jekyll/jekyll-redirect-from

JohnScottUK:
That's a pretty useful plugin. Thought many plugins not available in GitHub pages, but that one is in the list at https://pages.github.com/versions/

JohnScottUK:
Don't take this as the redirect.html layout interferes with jekyll-redirect-from (it uses it's own internal layout).

PR #3137, opened 2021-08-29
[Small Change] Fix repo details for utterances commenting

aidan-gallagher:
This is a bug fix.

## Summary

Make utterances find the repo text from the YAML config.

### DIFF
--- a/_includes/comments-providers/utterances.html
+++ b/_includes/comments-providers/utterances.html
@@ -10,7 +10,7 @@
 
     var script = document.createElement('script');
     script.setAttribute('src', 'https://utteranc.es/client.js');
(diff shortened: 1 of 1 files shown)


iBug:
I don't think this "fixes" anything. [The documentation][1] says you should provide `repository` in your `_config.yml`. Could you share some details on the reason for this change?

  [1]: https://mmistakes.github.io/minimal-mistakes/docs/configuration/#utterances-comments

aidan-gallagher:
Yes, you're right. 
I missed that when skimming through the documentation. 
I was following the instructions from others sources that I now realize were using Jekyll but a different theme.
I'm happy for this PR to be closed.

PR #3204, opened 2021-10-16
Update Brazilian Portuguese translation

georgeraraujo:

This pull request translates untranslated source texts and updates translations for a few translated source texts for Brazilian Portuguese (pt-BR).


mmistakes:
@georgeraraujo Thanks for adding in the missing translations.

Can you explain your motivation for changing some of the other texts? As someone who doesn't natively speak Brazilian Portuguese I can't determine if the new text is better than the previous.

I often get pull requests like this and I don't want to merge it in just to get another PR from someone else that preferred the previous translated texts.

PR #3326, opened 2022-01-08
404 <- include search options

manifestoso:

## Summary

Include search option upon 404 page loading.

### DIFF
--- a/docs/_pages/404.md
+++ b/docs/_pages/404.md
@@ -7,21 +7,6 @@ permalink: /404.html
 
 Sorry, but the page you were trying to view does not exist. --- perhaps you can try searching for it below.
 
-# <script async src="https://cse.google.com/cse.js?cx={{ site.google.search_engine_id}}"></script>
-
-{%- assign search_provider = site.search_provider | default: "lunr" -%}
-{%- case search_provider -%}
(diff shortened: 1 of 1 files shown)


manifestoso:
@mmistakes This PR has been tested to work for `google`. `algolia` is a bit buggy, where we will need to click the `.search__toggle` twice when 404 page is loaded to display the search box. This looks like is something to do with 

```
// Starting the search only when toggle is clicked
$(document).ready(function () {
  $(".search__toggle").on("click", function() {
    if(!search.started) {
      search.start();
    }
  });
});
</script>
```
in 
```
minimal-mistakes\_includes\search\algolia-search-scripts.html
```

Any thoughts?

mmistakes:
I think this is unnecessary. The 404 page included in the demo site is just that, a demo.
The theme ships with with a `search` layout which can be applied instead of needing to add the include to the page.

See below for an example:

```
---
title: "Page Not Found"
layout: search
excerpt: "Page not found. Your pixels are in another canvas."
sitemap: false
permalink: /404.html
---

Sorry, but the page you were trying to view does not exist --- perhaps you can try searching for it below.
```

Screen Shot 2022-01-10 at 8 08 36 AM

PR #3463, opened 2022-02-17
commit

minjoo95:


mmistakes:
@minjoo95 FYI you need to change the remote repository you keep pushing this commits to. It should point to your Git repo and not mmistakes/minimal-mistakes.

https://docs.github.com/en/get-started/getting-started-with-git/managing-remote-repositories

iBug:
@mmistakes Out of curiosity, how do you know one's pushing to the wrong repo, other than them keeping on submitting wrong PRs?

mmistakes:
It came up before in the past and that was the fix. I seem to remember someone noticed it when pushing their commits and not specifically opening a pull request.

The fact that it's so common with forked repos has me thinking there's a guide somewhere that new users to MM are following to get up and running and it's botching something.

That or it's just an unfamiliarity with how git works.

mmistakes:
It could also be due to how easy GitHub tries to make submitting pull requests after pushing up a new branch. I don't often work that way but feel like there's a big green open pull request button that shows up on github.com encouraging you.

Which if you're new to git seems like something you need to press.

PR #3648, opened 2022-05-11
Improve image quality when downscaling

DarkoGNU:
This is an enhancement.

## Summary

There's an issue with this theme, at least on Chrome. When the browser has to downscale an image, the quality is bad - the downscaled image is blurry.

This is a quick fix that dramatically improves the quality of downscaled images, at least on Chrome. I can also test it on other browsers if that's necessary.

## Test results

Note: you may need to download these images or open them in a separate tab to see the difference clearly.

#### Before

MenuBefore

#### After

MenuAfter

#### Before

AnimeBefore

#### After

AnimeAfter

### DIFF
--- a/_sass/minimal-mistakes/_reset.scss
+++ b/_sass/minimal-mistakes/_reset.scss
@@ -120,6 +120,7 @@ img {
   vertical-align: middle;
   border: 0;
   -ms-interpolation-mode: bicubic;
+  image-rendering: -webkit-optimize-contrast;
 }
 
 /* Prevent max-width from affecting Google Maps */


antenore:
That will work only on WebKit based browser. 
Did you tested it with `image-rendering: high-quality;` ?

Beside that I don't know if this should be set by default, maybe it's more something to let users customize by themselves

@mmistakes ?

DarkoGNU:
`high-quality` doesn't work on Chrome.

image

I've tested basically every possible flag, only `-webkit-optimize-contrast` improves the quality on Chrome. Maybe `high-quality` (or another flag) would provide similar results on browsers that aren't WebKit-based. I'll do some testing soon

mmistakes:
Not sure on this one. Seems like it could be a bit opinionated and have unintended consequences for certain types of images e.g., pixel art. 

https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering

DarkoGNU:
So… I've tested my change on:

### Firefox (Gecko)

Yeah, `-webkit-optimize-contrast` doesn't make a difference, but any option that makes a difference (like `optimizequality`) makes the image quality worse.

### Epiphany (WebKit)

`-webkit-optimize-contrast` makes a difference... and it's a bad difference. The image becomes too sharp

So, with that change, image quality (subjectively) improved on Chrome, doesn't change on Firefox (which gets it right without my change) and becomes worse on Epiphany (which also gets it right without my change).

Yeah, you're right that it's not a simple fix to merge & forget. I'll also try on mobile Chrome to see if there's a difference

DarkoGNU:
OK, `-webkit-optimize-contrast` makes a difference with Chrome on Android. The result?

Well, I'd say that the quality is improved. Look at this comparison ([source](https://danbooru.donmai.us/posts/5343407?q=incredibly_absurdres))

Android

The original image was downscaled using GIMP with Cubic interpolation.

The default rendering on Android's Chrome introduces some weird... I'm not sure, blockiness?

With `-webkit-optimize-contrast`, the image is smoother, more like the version downscaled with GIMP.

DarkoGNU:
OK, I have to admit - it seems to be a bug only with Chrome on Linux. I've checked on Windows - voilà, my images look fine without my changes, just like they look fine on Firefox.

Hell, it might even somehow be a bug specific to Arch Linux. There's no way of knowing unless I test this on other Linux distros.

I'm sorry, @mmistakes, for not checking whether this is OS-related before opening the pull request. I seriously didn't expect that 😮

DarkoGNU:
Well, the "downscaling bug" occurs even on a clean Chromium instance on my Arch Linux machine (I installed Chromium just to test that), but it doesn't occur on a Fedora machine. What a weird bug…

PR #3666, opened 2022-05-28
Remove IE9 upgrade notice

iBug:

## Summary

Since #3042 has been merged, the notice block is also only bloated bytes now.


dadahcr2:
Removing this file has resulted in Jeckyl failing builds

iBug:
@dadahcr2 If you have customized `default.html`, you should have removed that line by yourself, too.

PR #3669, opened 2022-05-29
Fix #3668 breaking "disable per-page when globally enabled"

iBug:
This is a bug fix.

## Summary

#3668 broke the original intent of "allow disabling breadcrumbs on a per-page basis when it's globally enabled (site-enabled)". I didn't realize that `!= false` held this intention.

Unfortunately [the `default` filter](https://shopify.github.io/liquid/filters/default/) doesn't fit this purpose (it eats `null`, `false` and empty strings) so I had to use another conditional. This should make both #3096 and the comment happy.

### DIFF
--- a/_layouts/single.html
+++ b/_layouts/single.html
@@ -8,7 +8,8 @@ layout: default
   {% include page__hero_video.html %}
 {% endif %}
 
-{% if page.url != "/" and site.breadcrumbs or page.breadcrumbs %}
+{% assign breadcrumb_enabled = page.breadcrumbs | default: site.breadcrumbs %}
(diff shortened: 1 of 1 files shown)


mmistakes:
Thanks @iBug. I actually tested this PR this time 😝  and it appears to work as intended.

- `breadcrumbs: false` in _config.yml disables them globally
- `breadcrumbs: true` in _config.yml enables them globally
- `breadcrumbs: false` in page's front matter disables them for only that page
- `breadcrumbs: true` in a page's front matter enables them for only that page (even when they aren't globally enabled in _config.yml)

PR #3704, opened 2022-06-28
Exclude lunr-en.js from Lunr search index

JMuff22:
This is a bug fix.

## Summary

Added the `search: false` to `lunr-en.js` to exclude it from the Lunr search index when enabling "search_within_pages" in the configuration file. 

## Context

This fixes a bug which was occurring in my own deployment of Minimal Mistakes where searching for specific terms was returning the `lunr-en.js` file. Searches now correctly return the right list based on search term. 

Other files may also need changing as well as this file if they are *never* needed in the search box e.g all files in `assets/`.

### DIFF
--- a/assets/js/lunr/lunr-store.js
+++ b/assets/js/lunr/lunr-store.js
@@ -1,5 +1,6 @@
 ---
 layout: none
+search: false
 ---
 
 var store = [


iBug:
I sort of feel like this patch should be applied to all `lunr-*.js` files. Can you verify if that causes any negative impact? If not would you apply to all such JS files?

JMuff22:
This shouldn't cause any negative impact to searching. 

`lunr-gr.js`- No problems searching in English and doesn't show up in searches. Cannot test Greek letters though. 

`lunr-en.js` - No problems searching in English and doesn't show up in searches
`lunr-store.js`- No problems searching in English and doesn't show up in searches

I would apply to all such JS files and if someone who is Greek or uses a website with Greek letters in could test that would be great.

iBug (approved):

ultimape:
I'm also getting sitemap.xml in the search, and possibly others. I tried adding these to the exclude: in _config.xml

Is there a way to avoiding needing to add this to the frontmatter of every while I don't want indexed?

JMuff22:
I resolved sitemap.xml and feed.xml from showing up in the search by commenting out these jekyll plugins. See https://github.com/FiQCI/dev/commit/b4da2aaa7c002861fbc4ca43f1f8752098af951f

PR #3765, opened 2022-07-10
⬆️ load latest fontawesome package

marciska:

## Summary

Upgrade to the newest `fontawesome` package.
Until now the package has been fixed at version `5.0.0` and has thus never received any bugfixes.
By changing the `@5` to `@latest` in the version tag, minimal-mistakes now ships always with the newest up-to-date version.
(Currently, `6.1.1` is the newest)

## Context

I am a researcher and therefore active on many research-related social platforms. I needed some icons that were introduced in later `patch-versions` of fontawesome version `5`, so I had to manually overwrite the `head.html` just to get the newest version. I believe that minimal-mistakes should always ship with the newest patch-version, as `5.0.0` is very outdated and does not include any patches that came after.
I am aware that minimal-mistakes should be kept minimal (and therefore not all requests can be met that only target specific individuals), but I assume that this patch should come in handy for a large-user base as there is no reason to limit ourselves to an old version if we can dynamically load always the latest release with `@latest`.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -15,8 +15,8 @@
 
 <!-- For all browsers -->
 <link rel="stylesheet" href="{{ '/assets/css/main.css' | relative_url }}">
-<link rel="preload" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@5/css/all.min.css" as="style" onload="this.onload=null;this.rel='stylesheet'">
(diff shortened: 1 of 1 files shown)


iBug:
FYI, the front page of [jsDelivr](https://www.jsdelivr.com/) shows that `@3` means any version with major=3. So the current implementation does automatically load the latest 5.x.y version of Font Awesome. It does not, however, load any FA v6.

Issue #997, opened 2017-05-06
Pagination link broken


wvengen:
When I'm using

```yaml
paginate_path: /pages/:num/
```

the pagination link for the first page appears to be broken, for example see [page 3](https://foodcoops.github.io/page/3/), where page two is referenced correctly, but the first page shows as `/page/:num/`. Sounds a like a bug - or did I do something unsupported in the [configuration](https://github.com/foodcoops/foodcoops.github.io/blob/master/_config.yml#L149)?

mmistakes:
What you have should work, not sure why it's not.

What happens if you remove `/` and do this in your `_config.yml` instead?

```
paginate_path: /page:num/
```

wvengen:
Thanks for your quick reply! It does work when I accept your suggestion (though I'm trying to get it working with the slash).

mmistakes:
Think you're out of luck then as it appears to be a limitation with Jekyll's pagination plugin. All of their examples use `page:num` vs what you want `page/:num`.

The plugin is no longer being developed so not sure if filling a bug there will have any effect. 

https://github.com/jekyll/jekyll-paginate

As an alternative Jekyll Paginate v2 is pretty great and more feature filled but it doesn't work with GitHub Pages unless you're building locally and pushing the built`_site` folder instead of having GH build.

https://github.com/sverrirs/jekyll-paginate-v2

wvengen:
Ah, sorry to bother you. Thanks!

mmistakes:
No worries.

PR #3778, opened 2022-07-14
Fix wrong `first_page_path` because of hard-coded `page:num`

yusanshi:
This is a bug fix.

## Summary

In code `assign first_page_path = ... | replace: 'page:num', '' | ...`, `page:num` is hard-coded. So if `site.paginate_path` doesn't ends with `page:num/` or `page:num`, `first_page_path` will get a wrong value.

The PR tries parsing the last part from `site.paginate_path`, instead of using a hard-coded `page:num`.

## Context

The same issue with: https://github.com/mmistakes/minimal-mistakes/pull/2431#issuecomment-598026992

> If `:num` is used alone (instead of `page:num`) then the previous code would generate wrong results. I have `paginate_path: /blog/:num/` in my config and I expect pages to have URLs like `/blog/2/` instead of `/blog/page2/`, which wasn't available before.

The author of  #2413 addressed the issue with Paginate V2. But I'm using GitHub Actions and only V1 is supported so I have to make the fix.

<br>

BTW, I'm new to Liquid and `"/blog/:num/" | split: "/"` returns `["", "blog", ":num"]`, instead of `["", "blog", ":num", ""]` as I expected. Weird... https://github.com/Shopify/liquid/issues/862

### DIFF
--- a/_includes/paginator.html
+++ b/_includes/paginator.html
@@ -1,6 +1,7 @@
 {% if paginator.total_pages > 1 %}
 <nav class="pagination">
-  {% assign first_page_path = paginator.first_page_path | default: site.paginate_path | replace: 'page:num', '' | replace: '//', '/' | relative_url %}
+  {% assign paginate_path_last = site.paginate_path | split: '/' | last %}
+  {% assign first_page_path = paginator.first_page_path | default: site.paginate_path | replace: paginate_path_last, '' | replace: '//', '/' | relative_url %}
   <ul>
     {% comment %} Link for previous page {% endcomment %}
     {% if paginator.previous_page %}


yusanshi:
Yes I meant "GitHub Pages". Thank you for pointing that out.

I searched with other keywords and finded the same issue #997.


As for Shopify/liquid#862, the  `split` filter in Liquid underlyingly calls `split` method of Ruby strings:
https://github.com/Shopify/liquid/blob/1fdc577246f400265532c6d9593e7323641ef403/lib/liquid/standardfilters.rb#L252-L261

I have not learned Ruby so I searched on Google and found it should be an expected behavior with Ruby.  The `split` method in Ruby has a second parameter `limit` and:
> If the limit parameter is omitted, trailing null fields are suppressed. If negative, there is no limit to the number of fields returned, and trailing null fields are not suppressed.  (From [Ruby doc](https://rubydoc.info/stdlib/core/String:split))

Liquid `split` accept only one parameter and `limit` is omitted. That's the cause of the weird behavior. Passing `limit=-1` will make it behave the same as other languages generally do.  See [SO](https://stackoverflow.com/questions/5453051/how-do-i-avoid-trailing-empty-items-being-removed-when-splitting-strings).

So I believe if Shopify/liquid#862 is to be fixed, they will add an optional `limit` parameter, not breaking what it currently is.

yusanshi:
Anyway, changing `/page:num/` to `/:num/` or `/page-:num/` should be a corner case. Closing this PR is OK to me. Someone in need will search, find the PR and get a solution.

iBug (approved):

PR #3958, opened 2022-10-19
Remove overlay and revert X to hamburger icon when popup disappears

ddomdel:
This is a bug fix.

## Summary

image

The pop up menu above disappears after you move the cursor out of it. However, the white overlay on the screen does not disappear, as well as the x nav button on the top-right, remains as is.

Now when you click anywhere on the screen, the popup appears again, and the x nav button turns back to a hamburger menu button.
And when you click on the hamburger menu, the popup disappears.

The functionality is now behaving in the exact opposite that we want to.

To remedy this, once the popup disappears the x nav button should also return back to normal (hamburger button)

This commit fixes this issue.

## Context

This is a direct pull request. No related Github issue


iBug:
No changes?

PR #4034, opened 2022-12-06
Add Docker Support

pourmand1376:

## Summary
I've added docker support.

## Context

With this commit, you no longer have to install ruby and run `bundle install` and install all dependencies if you want to run the image locally. You can just install docker and run:

```
docker-compose up
```
It will automatically download jekyll image and then install all required dependencies in an isolated environment and run the website from there.

I've already tested this and it is working as expected.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM jekyll/jekyll
+Label MAINTAINER Amir Pourmand
+#install imagemagick tool for convert command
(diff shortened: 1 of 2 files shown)


iBug:
Seems like a good idea. However the benefit for a bundled Docker setup is too narrow as it's only useful for theme development, so I'm not considering this PR for now.

PR #1657, opened 2018-05-08
Update example Reddit social share interpolation syntax

tlindsay42:
Resolves #1656

Issue #1656, opened 2018-05-08
Example Reddit social share URL interpolates relative URL rather than encoded absolute URL


tlindsay42:
## Description

The custom Reddit social sharing link example on the [Layouts](https://mmistakes.github.io/minimal-mistakes/docs/layouts/#social-sharing-links) page of the documentation specifies to populate the url parameter with `{{ page.url | relative_url }}`, which fails to populate the URL field in the popup window when implemented.  If the interpolation syntax is replaced with the encoded absolute value: `{{ page.url | absolute_url | url_encode }}`, as implemented in the theme default [social-share.html](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/social-share.html) file, the URL field will populate properly in the popup window.

```html
<a href="https://www.reddit.com/submit?url={{ page.url | relative_url }}&title={{ page.title }}" class="btn" title="{{ site.data.ui-text[site.locale].share_on_label }} Reddit"><i class="fab fa-fw fa-reddit" aria-hidden="true"></i><span> Reddit</span></a>
```

## Environment

- Minimal Mistakes version: 4.11.2
- Ruby gem or remote theme version: Remote theme
- Jekyll version: 3.7.3
- Git repository URL: https://github.com/troylindsay/troylindsay.github.io
- Operating system: Windows 10
- GitHub Pages hosted (if yes provide URL to site): https://www.troylindsay.io

---

## Expected behavior

The URL field should be populated with the page URL in the share on Reddit popup window.

## Steps to reproduce the behavior

1. Create an override social-share.html file and populate it with the code found [here](https://mmistakes.github.io/minimal-mistakes/docs/layouts/#social-sharing-links) page of the documentation.

```html
<a href="https://www.reddit.com/submit?url={{ page.url | relative_url }}&title={{ page.title }}" class="btn" title="{{ site.data.ui-text[site.locale].share_on_label }} Reddit">
  <i class="fab fa-fw fa-reddit" aria-hidden="true"></i><span> Reddit</span>
</a>
```

1. Serve the site
1. Navigate to a page with sharing enabled, and click on the Reddit link and the URL will not be populated.

Pull Request following momentarily.

PR #4066, opened 2022-12-31
Update Social Sharing links section

jstrong013:

## Summary
Update social sharing section for appropriate Reddit example and resource locations.

## Context
Could use a second set of eyes. Open to edits. 

Related to Issue #1656 and pull request #1657


iBug (changes requested):
LGTM except for the unnecessary `onclick` attribute. Might be better to keep it simple in the documentation.

Also `default: 'Share on'` is not required. There's more to break if `site.locale` is nonsensical.

PR #4114, opened 2023-01-23
Add rel="me" to footer links for web sign-in

infotexture:
✨ This is an enhancement or feature that builds on the microformat contributions by @dltj in https://github.com/mmistakes/minimal-mistakes/pull/3052.

Adding `rel="me"` to footer links [enables web sign-in](https://indieweb.org/How_to_set_up_web_sign-in_on_your_own_domain) or [RelMeAuth](https://microformats.org/wiki/RelMeAuth), so you can use your domain to sign in to other sites.

In https://github.com/mmistakes/minimal-mistakes/pull/3052, the `rel="me"` attributes were added to the _author profile_, but that only appears in post sidebars, and web sign-in requires these profile links to be present on the home page.

Adding this to the footer links ensures that these values are correctly detected by systems that check for the presence of bidirectional links between the author's home page and other profiles.

✅ Tested with the [Web Sign In](https://indiewebify.me/validate-rel-me/) test at https://indiewebify.me.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -7,7 +7,7 @@
     {% if site.footer.links %}
       {% for link in site.footer.links %}
         {% if link.label and link.url %}
-          <li><a href="{{ link.url }}" rel="nofollow noopener noreferrer"><i class="{{ link.icon | default: 'fas fa-link' }}" aria-hidden="true"></i> {{ link.label }}</a></li>
+          <li><a href="{{ link.url }}" rel="nofollow noopener noreferrer me"><i class="{{ link.icon | default: 'fas fa-link' }}" aria-hidden="true"></i> {{ link.label }}</a></li>
         {% endif %}
       {% endfor %}
     {% endif %}


dltj:
I'm curious — does the <code>rel="me"</code> in the footer operate differently than the one that is in the [author-profile.html](https://github.com/mmistakes/minimal-mistakes/pull/3052/commits/ba550298e153c5346424e416429ea0aa675271f3#diff-4c101b945212d26f80e9806e855f0230d9d866d068b816f2446654f5fe88e84eR50)

infotexture:
> I'm curious — does the `rel="me"` in the footer operate differently than the one that is in the [author-profile.html](https://github.com/mmistakes/minimal-mistakes/pull/3052/commits/ba550298e153c5346424e416429ea0aa675271f3#diff-4c101b945212d26f80e9806e855f0230d9d866d068b816f2446654f5fe88e84eR50)

No, but as described above,

> that only appears in post sidebars, and web sign-in requires these profile links to be present on the home page.

dltj:
Ah, I think I see...a critical difference between your site and mine is that you have overridden the homepage template so the `author-profile.html` file isn't included. Makes sense. This is a good change.

iBug:
I don't think it would make sense to unconditionally add `rel="me"` to every footer link. Perhaps an option for each link item would be better.

(Contrary to Michael, I'm neutral to adding more config options as they're a requisite of flexibility.)

lsolesen:
Here is the suggested pull request: https://github.com/mmistakes/minimal-mistakes/pull/5030

lsolesen:
@iBug for some reason my pull request got flagged like spam. I have no idea why? https://github.com/mmistakes/minimal-mistakes/pull/5030

PR #4143, opened 2023-02-06
Google analytics V4

derek-adair:

## Summary
Google is deprecating google-universal in July.  They are encouraging to swap to V4, which includes a new tag.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -96,7 +96,7 @@ social:
 
 # Analytics
 analytics:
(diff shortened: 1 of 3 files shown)


iBug:
Is this any different from [`google-gtag`](https://github.com/mmistakes/minimal-mistakes/blob/master/_includes/analytics-providers/google-gtag.html)?

derek-adair:
my bad it would appear not!

PR #4436, opened 2023-07-25
Renamed navigation item to docs and samples

onefloid:
## Summary

I searched for the docs and found this great documentation under Sample Collection. It was not intuitive. So I suggest renaming it to Docs & Samples.  I did not change the url to avoid broken links.


iBug:
Pretty much every link on the front page links to a documentation page, from where you can navigate through all docs pages from the left sidebar.

image

That documentations appear in the collections archive is simply because they belong to a collection named `docs`, and I'd rather consider this a coincidence than an intention.

PR #4459, opened 2023-08-05
Add Devcontainer for Jekyll build

wsciaroni:

## Summary

As a developer, I want to keep my projects' dependencies isolated during development.

## Context

This is a quality of life improvement for developers consuming this project.

## The What

Visual Studio Code is my IDE of choice when working FOSS projects. It has support for containerized development with very little configuration on the developer's side.

## How

To use this, simply install the following:

- VS Code
- Containerized Development Extension
- Docker

Then, open the container with VSCode. You should be prompted to "Reopen in a container".

This config automatically takes care of port forwarding for you.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,31 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/jekyll
+{
+	"name": "Jekyll",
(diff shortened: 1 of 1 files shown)


iBug:
Since neither Michael nor I am using devcontainers, I don't want to include the config into our repository as it puts maintenance on us (e.g. `bullseye` is now updated to `bookworm`). If you're a heavy user of devcontainers, you can keep the config in your local clone of the repo, and add it to `.git/info/exclude` so it won't be checked into the repo when you do `git add`.

Issue #2184, opened 2019-06-11
Use Font Awesome Kits?


coliff:
Font Awesome have changed the way they offer downloads and have a new implementation method which could be useful: https://blog.fontawesome.com/introducing-font-awesome-kits-7134d1d59959
By generating a custom link you can choose 'Latest' so you always get the latest version which has been released. (not sure what the cache life is like though).
Worth taking a look and deciding if it's something worth doing: https://fontawesome.com/

mmistakes:
Nice. I went ahead and created a kit and as far as I can everything looks good. Going to push the change up to `master` and see how things go.

Should hopefully save you from having to PR the newest version of FA and keep it current. 🤣

Issue #3649, opened 2022-05-13
The YouTube video on the "Layout: Header Video" page is private


bewuethr:
### Motivation

The [example post with a header video](https://mmistakes.github.io/minimal-mistakes/layout/uncategorized/layout-header-video/) ([source](https://github.com/mmistakes/minimal-mistakes/blob/master/docs/_posts/2010-06-02-post-video-youtube.md)) looks like this:

![image](https://user-images.githubusercontent.com/8521043/168209010-20b7e834-05db-4cab-893b-9515e4559203.png)

i.e., doesn't demonstrate what a header video would look like.

### Suggestion

Use a non-private video 🙂

bewuethr:
Interestingly, when I visit the link from the source directly, I can access the video 🤔

mmistakes:
Not a theme issue... I'm guessing the video owner has turned off embedding/sharing and that's why it doesn't show up anymore.

Switching the embeds to public videos...

PR #4475, opened 2023-08-14
Update to font-awesome v6 (more icons)

kangsterizer:
## Summary

Update to font-awesome v6 (more icons)

## Context

I saw that in the past you've used [kit.fontawesome.com](https://kit.fontawesome.com/4eee35f757.js) (issue #2184) but then have reverted to the hard-coded version. Not sure if this should instead use the kit feature again, but in the mean time - updating to v6 allows using all the free icons documented on fontawesome's website. These are available on the linked CDN and `@fortawesome`.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -15,8 +15,8 @@
 
 <!-- For all browsers -->
 <link rel="stylesheet" href="{{ '/assets/css/main.css' | relative_url }}">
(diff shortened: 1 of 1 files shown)


magic-happenz:
Could you maybe implement it in a way the source is customizable? Loading fonts is not compatible with GDPR which makes the template unusable in all Europe.

srpomeroy:
Nice one @kangsterizer. I added your change to my custom head file and am finally able to display the `fa-x-twitter` icon.

iBug:
I consider this one superseded by #3765

mmistakes:
Closing, #3649 was merged which loads in the latest FA.

PR #4679, opened 2024-01-31
Using locale & language tags from page instead of site

alloystorm:

## Summary

By using the locale tag from the page, it allows different language pages on the same site. 

## Context

Previously it only allows a site to have only one language, which limits the ability to create multi-lingo sites. 

With this proposed change, the pages are rendered using the locale tag from the page front matter, so that you can create pages for different languages on your website.

### DIFF
--- a/_includes/author-profile.html
+++ b/_includes/author-profile.html
@@ -23,7 +23,7 @@
   </div>
 
   <div class="author__urls-wrapper">
-    <button class="btn btn--inverse">{{ site.data.ui-text[site.locale].follow_label | remove: ":" | default: "Follow" }}</button>
+    <button class="btn btn--inverse">{{ site.data.ui-text[page.locale].follow_label | remove: ":" | default: "Follow" }}</button>
     <ul class="author__urls social-icons">
       {% if author.location %}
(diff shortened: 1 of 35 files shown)


alloystorm:
I have my website running with this change. 

For example this page: https://vrstormlab.com/dancexr/
Click on one of the language links below the title and you'll land on a new page with the same content but for a different language. The locale and lang tags are set correctly for each page so that the browser can recognize.

iBug:
I don't think `page.locale` propagates from `site.locale`. This means existing users who don't have proper `page.locale` will suddenly find their site back in English. If the fallback to `site.locale` is taken care of, I'll be OK with this PR.

iBug:
Here's my new idea: Since all templates end up being rendered in `default.html`, we can assign our desired variables in `default.html` and use it everywhere. Something like:

```liquid
<!-- _layouts/default.html -->
{% assign locale = page.locale | default: site.locale %}

<!-- everywhere else -->
{{ site.data.ui-text[locale].blahblahblah }}
```

If you'd like to proceed, consider taking this approach or we can discuss another idea that you find better.

iBug:
Superseded by #5439.

PR #4862, opened 2024-05-30
Page locale

suhlig:
This is a feature. 

## Summary

This PR allows for a page to specify a content `locale` that is used only for _that one_ page.

Compared to the similar PR https://github.com/mmistakes/minimal-mistakes/pull/4679,  this set of changes is relatively small, as I am not going for localising anything. Instead, the focus is on expressing the _actual_ locale of the content, as it is written in the page and declared as `locale` in the front matter, in the containing element's `lang` attribute.

Fallback to `site.locale` is present where applicable.

This PR has an associated commit that writes the site locale (`lang` attribute of the `html` element) in the style the W3C recommends (language and region separated by dash), instead of just cutting off the region. Let me know if you'd prefer this one as a separate PR.

## Context

I am running [my blog](https://uhlig.it) mostly in English, but a few pages are in German. I am running a spell checker across all pages before publishing them, and having an accurate locale on each page is crucial for that.

## Example

A site that has `locale: en_US` in its `_config.yaml` generates the following `html` element:

```html
<html lang="en-US" ...>
```

A page may specify that its content is written in German:

```markdown
---
title: Eine Seite in Deutsch
locale: de_DE
---

Das ist der Inhalt...
```

With this PR, the rendered HTML looks like this:

```html
<html lang="en-US"...>
<article class="page" ... lang="de-DE">
  <h1>Eine Seite in Deutsch</h1>
  ...
```

Everything is still being treated as written in the site's locale `en-US`, but this one page declares its main content correctly as `de-DE`.

### DIFF
--- a/_includes/archive-single.html
+++ b/_includes/archive-single.html
@@ -11,7 +11,7 @@
 {% endif %}
 
 <div class="{{ include.type | default: 'list' }}__item">
-  <article class="archive__item" itemscope itemtype="https://schema.org/CreativeWork"{% if page.locale %} lang="{{ post.locale | default: site.locale | replace: "_", "-" | default: "en" }}"{% endif %}>
+  <article class="archive__item" itemscope itemtype="https://schema.org/CreativeWork"{% if post.locale %} lang="{{ post.locale }}"{% endif %}>
     {% if include.type == "grid" and teaser %}
       <div class="archive__item-teaser">
         <img src="{{ teaser | relative_url }}" alt="">
--- a/_includes/archive-single.html
+++ b/_includes/archive-single.html
@@ -11,7 +11,7 @@
 {% endif %}
 
 <div class="{{ include.type | default: 'list' }}__item">
-  <article class="archive__item" itemscope itemtype="https://schema.org/CreativeWork" lang="{{ post.locale | default: site.locale | replace: "_", "-" | default: "en" }}">
+  <article class="archive__item" itemscope itemtype="https://schema.org/CreativeWork"{% if page.locale %} lang="{{ post.locale | default: site.locale | replace: "_", "-" | default: "en" }}"{% endif %}>
     {% if include.type == "grid" and teaser %}
       <div class="archive__item-teaser">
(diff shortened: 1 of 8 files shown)


iBug:
Nice job. Can this be changed so that when a page-level override isn't present, it won't emit the `lang` attribute at all?

suhlig:
Hmm, I did not think about it that way. My thought was that we _always_ know the language (at least the default system fallback would be available).

But if it's a blocker to merging the PR I can certainly make that change.

iBug:
My take on this is that it's a per-page settings override, so the `lang` tag need not be present if nothing was overridden. A secondary reason is to keep the changes at a minimal level for existing users, such that the output HTML doesn't change unnecessarily across theme versions.

suhlig:
Makes sense. I'll update the PR.

suhlig:
I pushed another commit that emits the lang attribute only if `page.locale` is set. I marked it as fixup so that it stays separate while we are discussing it, but at rebase time it should probably be squashed with the previous one.

_includes/archive-single.html
```diff
@@ -11,7 +11,7 @@
 {% endif %}
 
 <div class="{{ include.type | default: 'list' }}__item">
-  <article class="archive__item" itemscope itemtype="https://schema.org/CreativeWork">
+  <article class="archive__item" itemscope itemtype="https://schema.org/CreativeWork"{% if page.locale %} lang="{{ post.locale | default: site.locale | replace: "_", "-" | default: "en" }}"{% endif %}>
```
iBug, line 14: With `{% if %}` guarding the presence of the per-post locale setting, all the `default`s can be safely dropped.

Specific to this one: Did you mix up `page` and `post`?

iBug (changes requested):

_includes/archive-single.html
suhlig: I had the exact same thought five seconds after pushing ;-) I'll update the PR.

suhlig: Fixed In b5fd4dae

Issue #3903, opened 2022-09-10
RTL layout for fa-IR and ar-AR localisations


srezasm:
### What happened?

- Minimal Mistakes version: 4.24.0
- Ruby gem or remote theme version: Any
- Jekyll version: 3.9
- Git repository URL: https://github.com/srezasm/blog
- Hosted on GitHub Pages (if yes provide URL to site): https://srezasm.github.io/blog/
- Operating system: Linux

### Expected behavior

The theme must adjust to the RTL layout by choosing a right-to-left language localization like fa-IR and ar-AR

### Steps to reproduce the behavior

1. Cloned the main repo
2. Configured the GitHub pages settings

srezasm:
Here you can see a RTL post example: https://srezasm.github.io/blog/rtl/rtl-post

I changed every "right" with "left" and vice versa in the inspector > sources > main.css and added `direction: rtl` to the body element and here is the result:
![screencapture-srezasm-github-io-blog-rtl-rtl-post-2022-09-10-18_38_39](https://user-images.githubusercontent.com/45644975/189487131-740e4ac9-67a2-4d67-9662-37ec876abe6e.png)

N013M:
can you tell me how did you do it?

srezasm:
> can you tell me how did you do it?

Here is the link to my RTL fork:
https://github.com/srezasm/minimal-mistakes-rtl

iBug:
Implemented in #4886.

PR #4884, opened 2024-06-21
feat: add rtl support

ha36d:
This is an enhancement or feature.

## Summary

It will add RTL support for languages like Arabic, Persian, etc.

## Context

Is this related to any GitHub issue(s)? yes, Closes mmistakes/minimal-mistakes#3903

PR #4886, opened 2024-06-21
Add RTL Support

yaim:

## Summary

As you might know, some writing systems are [written and read from right to left](https://en.wikipedia.org/wiki/Right-to-left_script) (mainly Persian, Arabic & Hebrew).  These types of writings need extra formatting for web.

In this PR I added the support for RTL direction with the minimum changes in the code by only adding an extra css file, & not changing the html structures and leaving zero side effect for the original left to right view.

These changes are just meant to be proof of concept and are not final changes. I was not sure if RTL support is a concern for the project or not, but if it is, then I'll be more than happy to make them ready for a final review.

## Context

RTL support was previously mentioned in #3903 but the suggested changes where affecting the original LTR direction.

## Live Demo

Here's a [live demo](https://blog.yash.im/fa/open%20source/%D8%B3%D9%84%D8%A7%D9%85-%D8%AF%D9%86%DB%8C%D8%A7/) of the RTL support.


yaim:
@ha36d I noticed you have submitted a related PR (#4884), 15 minutes before me. 😅
I'll be glad if we can work on it together.

ha36d:
> @ha36d I noticed you have submitted a related PR (#4884), 15 minutes before me. 😅 I'll be glad if we can work on it together.

Yes, Sure. I was moving from blogger to jekyll. So I needed RTL support on this theme.

ha36d:
> Awesome! The changes I suggested are activated by adding a
> 
> ```yaml
> direction: ltr
> ```
> 
> to the site settings sections in `_config.yml`.
> 
> Here's the [config](https://github.com/yaim/blog-fa/blob/blog/_config.yml#L19) for the demo link mentioned in the PR description.

For me, it is the same. But I am not introducing css changes. It only makes the body page "RTL". I guess still we need the body direction in "rtl".

yaim:
Correct. I added the body direction to CSS, but I think adding the `dir` attribute is [more appropriate](https://www.w3.org/International/questions/qa-bidi-css-markup).
But the rest of the CSS extensions were needed to keep to meaning of the template.

ha36d:
> Correct. I added the body direction to CSS, but I think adding the `dir` attribute is [more appropriate](https://www.w3.org/International/questions/qa-bidi-css-markup). But the rest of the CSS extensions were needed to keep to meaning of the template.

Agreed. you can copy the documentation part from mine. I will close my PR.

yaim:
Thanks! Cherry picked your doc change! 🙌🏻

iBug:
Looks decent. I have one minor concern: For the configuration key, why did you choose `direction: rtl` instead of a boolean key like `rtl: true`? To me it seems easier to maintain and you don't have to mention the default value (which is actually `null`).

yaim:
Thanks for the feedback @iBug, that totally makes sense!
I also noticed the toc does not expand in mobile view which is not the as original version. I'll fix both soon and ask for your review.

Also I thought maybe it's better to inject a `rtl` value into body's attributes, instead of loading a whole new file and then have the rtl condition styles next to the parts their original style is defined. I think that would be easier to maintain and if someone decided to change the style, they would also notice there's a rtl counterpart. What do you think?

yaim:
Aaaaah, there's actually an exact example in the scss docs:

```scss
  // It can also be used to style the outer selector in a certain context, such
  // as a body set to use a right-to-left language.
  [dir=rtl] & {
    margin-left: 0;
    margin-right: 10px;
  }
```

yaim:
@iBug I pushed the change you mentioned + fix for the bug I saw on small screen view.

Also made the removing separate rtl file suggestion in the 3rd commit (50bd4eaff924ec49b551dffeaa0b10bb9f67ced3).
If you think it's OK, I'll fixup it to the first commit, but if you don't like that, I can drop it.

Thanks! 🙌🏻

iBug (changes requested):

iBug:
Seems sufficiently good for now so I've merged it. If you have any further improvements feel free to open another PR.

One potential improvement I can think of is replacing all affected margins and paddings with [`padding-block-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block-start) and similar, so that you don't have to, for example, set `*-left: 0` while `*-right: (original value)`. [Browser support](https://caniuse.com/mdn-css_properties_padding-block-start) for these direction-agnostic CSS properties are not to be worried.

iBug:
I've pushed ded2d87b2bddecabbd1f4c9d6b7558ec450a9dc9 and I wonder if you'd like to help test it (e.g. `remote_theme: "mmistakes/minimal-mistakes@ded2d87b2bddecabbd1f4c9d6b7558ec450a9dc9"`).

yaim:
Sorry for late reply.
All my test cases look valid on the changes! They look great! 🙌🏻

srezasm:
Hi, thanks for your work. It would be great to dynamically adjust the direction of the paragraph based on the used language.

Just like the way that GitHub handles RTL languages in rendered markdown views:

همانطور که مشاهده می‌کنید، جهت این پاراگراف از راست به چپ است.

iBug:
@srezasm If you use Kramdown's syntax for adding HTML attributes, you'll automatically have RTL for only specified paragraphs:

```text
This is some text.

This is some more text.
{: dir="rtl" }
```

srezasm:
@iBug Can it be automated by the language that paragraph is started with?

srezasm:
@iBug It works with dir="auto" as well, but still, I will need to add this at the end of each paragraph manually.

iBug:
@srezasm You can *probably* just add one `dir="auto"` to the `<body>` element as this attribute can be inherited.

srezasm:
@iBug Didn't work, do you know how to add it to all individual text placeholder tags?

PR #4958, opened 2024-08-19
Remove deprecated dependency jekyll-algolia

alouiadel:

## Summary
Removed mentioning of now deprecated jekyll-algolia gem from documentation
i also suggest you check : https://github.com/mmistakes/mm-github-pages-starter/pull/59
## Context
No, i had to setup my personal website using this template , this is how i encountered this.

### DIFF
--- a/docs/_docs/05-configuration.md
+++ b/docs/_docs/05-configuration.md
@@ -625,61 +625,6 @@ lunr:
 **Note:** Only the first 50 words of a post or page's body content is added to the Lunr search index. Setting `search_full_content` to `true` in your `_config.yml` will override this and could impact page load performance.
 {: .notice--warning}
 
-#### Algolia
-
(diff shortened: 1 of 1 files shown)


alouiadel:
It seems that no alternative gem is available since Jekyll isn't popular enough to encourage API providers to create these gems. However, I view keeping a deprecated gem active as a vulnerability and a security concern for the template users

iBug:
Since Algolia integration remains a stood-out feature of the theme, I prefer keeping the documentation or one of its substitutes in place, until the status quo stops working altogether.

Also, I don't consider a deprecated gem "a vulnerability" in a static site generator until proven to be dangerous. Deprecation only means it'll stop being maintained, not being taken over by hostile parties.

alouiadel:
ok fair enough

PR #4959, opened 2024-08-20
Fix: Prevent double title issue in SEO metadata by updating seo.html

alouiadel:
This is a bug fix.

## Summary
This pull request addresses the issue of duplicated titles in the SEO metadata. The fix involves overriding the seo.html file and     updating the logic to ensure the page title doesn't include the site title twice. The change ensures that the title displayed in the     browser tab and metadata is clean and correctly formatted.

Changes Include:
- Created a custom _includes/seo.html file.
- Updated the logic in seo.html to prevent the page title from containing the site title twice.

Testing:
- Verified that the title is no longer duplicated in the browser tab and in the generated HTML meta tags.

### DIFF
--- a/_includes/seo.html
+++ b/_includes/seo.html
@@ -2,7 +2,11 @@
 {%- assign title_separator = site.title_separator | default: '-' -%}
 
 {%- assign page_title = page.title | default: site.title | replace: '|', '&#124;' -%}
-{%- assign seo_title = page_title | append: " " | append: title_separator | append: " " | append: site.title | replace: '|', '&#124;' -%}
+{%- if page_title contains site.title -%}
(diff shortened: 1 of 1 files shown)


iBug:
I kind of consider it a feature that the page title uses `site.title` as a placeholder. And to keep in line with other pages where the title always ends in `- {{ site.title }}`, I'm not sure how a single page titled only `{{ site.title }}` would look better.

alouiadel:
Example after fixing it in my blog : 
image
its only site title  Adel Aloui Portfolio/Blog , instead of  Adel Aloui Portfolio/Blog |  Adel Aloui Portfolio/Blog 
image
but here site title  Adel Aloui Portfolio/Blog , gets added again

PR #5014, opened 2024-10-24
Added read_time_override

karmaniverous:

## Summary

If I create a post and then later move the content to a page in a collection, for continuity I want to keep the original post, but I'll forward it to the new page. Since the original post no longer has content, read time on the archive page will be < 1 min.

`read_time_override` allows you to set read time manually so the read time in the post listing can match the read time where the actual content is.

Example: see the read times of any of [these posts](http://localhost:4000/tags/#project-governance) under `project-governance`. The posts themselves are empty, just have front matter, and before this change all read times showed < 1 min. Click through to see the content in its new collection location.

## Context

No


README.md
```diff
@@ -1,3 +1,5 @@
+> This fork just offers a simple enhancement to post metadata. Here's [the PR](https://github.com/mmistakes/minimal-mistakes/pull/5014).
```
iBug, line 1: Is this line intended to come up in the PR?

iBug (changes requested):

README.md
karmaniverous: Ah no sorry! I'll kill it.

karmaniverous: reverted

PR #5021, opened 2024-10-30
nav_list empty links

karmaniverous:

## Summary

An empty-string child url in `nav_list` should produce an unlinked title.

## Context

By default, a `nav_list` child entry without a link defaults to a link to the current page.

There are times when what you really want is NO link. See the sidebar [here](https://karmanivero.us/projects/entity-manager/intro/) for an example. To maintain backwards compatibility, this change enables this behavior only when the `url` property is explicitly set to an empty string, like this:

```yml
entity-manager:
  - title: Packages
    children:
      - title: <strong><code>entity-manager</code></strong>
        url: '' # 👈 THIS!
      - title: <i class="fa-brands fa-github fa-lg"></i>&nbsp;&nbsp;GitHub
        url: https://github.com/karmaniverous/entity-manager
      - title: <i class="fa-solid fa-heart fa-lg"></i>&nbsp;&nbsp;NPM
        url: https://www.npmjs.com/package/@karmaniverous/entity-manager
      - title: <i class="fa-solid fa-book fa-lg"></i>&nbsp;&nbsp;API
        url: https://docs.karmanivero.us/entity-manager/
      - title: <i class="fa-solid fa-comment fa-lg"></i>&nbsp;&nbsp;Help
        url: https://github.com/karmaniverous/entity-manager/discussions
```


iBug:
What about replacing "empty URL" with either `null` or `false`, so that `{% if child.url == "" %}` can be written as `{% if child.url %}`? I also think these are more explicit than an empty string.

karmaniverous:
I was concerned about backwards compatibility. I think if you make that change a bunch of people my find their existing links no longer working as expected.

PR #5030, opened 2024-11-11
Change to Indieweb defaults for rel-me and rel-author

lsolesen:
This is a bug fix. 

## Summary

I think the current implementation of rel-me and rel-author is flawed. 

rel-me is supposed to be used on profile pages relating to other profile pages - https://microformats.org/wiki/rel-me

Right now it is used on posts from the author-block, but here rel-author should be used instead - https://microformats.org/wiki/rel-author

I added the possibility from the authors.yml to add rel to the individual links, if you want to add another rel. This should cater for everybody. However, people might need to update this, if they want to keep it.

I also added the possibility to add rel from footer links, if people want that.

## Context

Also I added h-entry, so Jekyll sites will be more compatible with Indieweb straight out of the box.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -7,7 +7,7 @@
     {% if site.footer.links %}
       {% for link in site.footer.links %}
         {% if link.label and link.url %}
-          <li><a href="{{ link.url }}" rel="nofollow noopener noreferrer"><i class="{{ link.icon | default: 'fas fa-link' }}" aria-hidden="true"></i> {{ link.label }}</a></li>
+          <li><a href="{{ link.url }}" rel="nofollow noopener noreferrer{% if item.rel %} {{ item.rel }}{% endif %}"><i class="{{ link.icon | default: 'fas fa-link' }}" aria-hidden="true"></i> {{ link.label }}</a></li>
         {% endif %}
       {% endfor %}
     {% endif %}
--- a/_includes/page__date.html
+++ b/_includes/page__date.html
@@ -1,6 +1,6 @@
 {% assign date_format = site.date_format | default: "%B %-d, %Y" %}
 {% if page.last_modified_at %}
-  <p class="page__date"><strong><i class="fas fa-fw fa-calendar-alt" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].date_label | default: "Updated:" }}</strong> <time class="dt-published" datetime="{{ page.last_modified_at | date: "%Y-%m-%d" }}">{{ page.last_modified_at | date: date_format }}</time></p>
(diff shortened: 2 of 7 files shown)


iBug:
Sorry for the confusion. Next time make sure you have read everything provided in the PR template. I added those checks because this repo is a victim of constant [misdirected PRs](https://github.com/mmistakes/minimal-mistakes/pulls?q=is%3Apr+is%3Aclosed).

_includes/page__date.html
```diff
@@ -1,6 +1,6 @@
 {% assign date_format = site.date_format | default: "%B %-d, %Y" %}
 {% if page.last_modified_at %}
-  <p class="page__date"><strong><i class="fas fa-fw fa-calendar-alt" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].date_label | default: "Updated:" }}</strong> <time class="dt-published" datetime="{{ page.last_modified_at | date: "%Y-%m-%d" }}">{{ page.last_modified_at | date: date_format }}</time></p>
+  <p class="page__date"><strong><i class="fas fa-fw fa-calendar-alt" aria-hidden="true"></i> {{ site.data.ui-text[site.locale].date_label | default: "Updated:" }}</strong> <time class="dt-updated" datetime="{{ page.last_modified_at | date: "%Y-%m-%d" }}">{{ page.last_modified_at | date: date_format }}</time></p>
```
infotexture, line 3: Good catch. I recall stumbling over this inconsistency a while back and implemented a local workaround, but makes sense to fix this at the source.

_includes/footer.html
```diff
@@ -7,7 +7,7 @@
     {% if site.footer.links %}
       {% for link in site.footer.links %}
         {% if link.label and link.url %}
-          <li><a href="{{ link.url }}" rel="nofollow noopener noreferrer"><i class="{{ link.icon | default: 'fas fa-link' }}" aria-hidden="true"></i> {{ link.label }}</a></li>
+          <li><a href="{{ link.url }}" rel="nofollow noopener noreferrer{% if item.rel %} {{ item.rel }}{% endif %}"><i class="{{ link.icon | default: 'fas fa-link' }}" aria-hidden="true"></i> {{ link.label }}</a></li>
```
infotexture, line 10: Might be worth explaining in the docs how this works, so people know what they need to do to extend the attribute values here.

infotexture (commented):
@lsolesen These changes look like good extensions to the microformats support here. 

I don't have a strong opinion on when `author` is more appropriate than `me`, but that distinction may be less relevant on single-author sites.

Curious though: would these changes still pass the [Web Sign In](https://indiewebify.me/validate-rel-me/) test at https://indiewebify.me, which seems to expect `rel="me"` and not `rel="author"`?

Found one minor nit in the placement of the `p-name` property.

PR #5033, opened 2024-11-11
Modern font stack

lsolesen:

## Summary

Provided in variables are now modern font stacks which can easily be used to switch the font of the site.

## Context

Added the font definitions from the modern font stack to be able to quickly switch fonts if desired.


iBug:
Not sure how much utility this has. Isn't it already possible to override variables locally?

shiami:
+1 for better management.

iBug (approved):

PR #5103, opened 2025-01-27
icon color: bluesky, instagram, threads, tiktok, x, youtube

codelabor:

## Summary

Positioned the brand icon below in alphabetical order

- youtube

Added color for the brand icon below

- blueskey

Modified the color for the brand icon below

- instagram

Added a square class to the brand icon below

- instagram

Color crashed for the brand icon below, so I consolidated it into one setting

- threads
- tiktok
- x

### DIFF
--- a/_sass/minimal-mistakes/_utilities.scss
+++ b/_sass/minimal-mistakes/_utilities.scss
@@ -201,13 +201,14 @@ $text-alignments: left, right, start, end, center, justify;
   @each $color, $icons in (
     $behance-color: ".fa-behance, .fa-behance-square",
     $bitbucket-color: ".fa-bitbucket",
+    $bluesky-color: ".fa-bluesky, .fa-square-bluesky",
     $dribbble-color: ".fa-dribbble, .fa-dribbble-square",
     $facebook-color: ".fa-facebook, .fa-facebook-square, .fa-facebook-f",
     $flickr-color: ".fa-flickr",
     $foursquare-color: ".fa-foursquare",
     $github-color: ".fa-github, .fa-github-alt, .fa-github-square",
     $gitlab-color: ".fa-gitlab",
-    $instagram-color: ".fa-instagram",
+    $instagram-color: ".fa-instagram, .fa-square-instagram",
     $keybase-color: ".fa-keybase",
     $lastfm-color: ".fa-lastfm, .fa-lastfm-square",
     $linkedin-color: ".fa-linkedin, .fa-linkedin-in",
@@ -223,6 +224,7 @@ $text-alignments: left, right, start, end, center, justify;
     $vine-color: ".fa-vine",
     $xing-color: ".fa-xing, .fa-xing-square",
     $youtube-color: ".fa-youtube",
+    $black-color: ".fa-x-twitter, .fa-square-x-twitter, .fa-threads, .fa-square-treads, .fa-tiktok",    
   ) {
     #{$icons} {
       color: $color;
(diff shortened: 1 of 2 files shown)


_includes/author-profile.html
```diff
@@ -22,6 +22,7 @@ <h3 class="author__name p-name" itemprop="name">
     {% endif %}
   </div>
 
+  {% if author %}
```
iBug, line 25: This change is unnecessary. We're already using `{% if author.avatar %}` above, so either the build errors out, or this check always passes.

_sass/minimal-mistakes/_page.scss
```diff
@@ -2,560 +2,565 @@
    SINGLE PAGE/POST
    ========================================================================== */
 
-#main {
```
iBug, line 5: Changes in this file are incomprehensible. Please revert.

iBug (changes requested):

_includes/author-profile.html
codelabor: Regardless of author.avatar's conditional statement.
The intention is to prevent the following phenomenon.

[Intended situation]
- I want to mark the avatar.
- I don't want to show links.

[Actual results]
- Avatar is displayed. (It's OK)
- Links are not displayed. (It's OK)
- However, when the screen is reduced, the 'Follow' button is displayed for no use.' (Why?)
- Even if I click the 'Follow' button, there are no links to display, so I only see empty spaces. (Why?)

[Conclusion]
In the end, the phrase {%ifauthor%} was used to turn on/off the "url wrapper" that determines the "follow" button display.
If there is a better way, it doesn't matter if you write another conditional statement. 

What I wanted to do was disable url wrapper because if there is no information to follow, such as links and sns accounts, there is no reason for the 'follow' button to be displayed.

_sass/minimal-mistakes/_page.scss
codelabor: Something seems a little strange.
It was the 92nd line I was going to add the code below.

    h2:first-of-type {
      margin-top: 0em;
    }

The reason why I need to put that setting in is as follows.

[AS-IS]
The height of the profile panel on the left and the TOC panel on the right are the same, but the height of the content panel in the middle is slightly lower. 
This is because the margin-top of the h2 tag is applied.
The margin-top of the h2 tag should be applied from the use of the second h2 tag.

[TO-BE]
The margin-top of h2 used for the first time on the intermediate content panel shall not be set.

iBug: But the `git diff` is still incomprehensible. Please revert the unnecessary part and make sure the diff accurately reflects your changes.

iBug: Also, `:first-of-type` might not be as appropriate as `:first-child`, in case the content starts with something else. For example, I usually start my blog article with a regular paragraph.

codelabor: Then how about this?

  /* vertical alignment for first content */
    :first-child, 
    aside + :nth-child(2) {
      margin-top: 0em;
  }

_includes/author-profile.html
iBug: I'm still not convinced this will work as intended. For `{% if author %}` to fail, you must omit the entire `author` key (so it evaluates to `nil`). In this case, there's no way to display an avatar.

_sass/minimal-mistakes/_page.scss
```diff
@@ -170,6 +170,12 @@ body {
     margin-top: -1.5em;
     padding-inline-start: 1.25rem;
   }
+
+  /* vertical alignment for first content */
+    :first-child, 
+    aside + :nth-child(2) {
```
iBug, line 176: These two lines are over-indented by 2 spaces (1 level).

PR #5161, opened 2025-03-12
Removed Exessive New Lines via Liquid

cocacool49:

## Context
When checking on the HTML source I noticed extensive new lines.
I have checked on the liquid and {% ... %} should be {%- ... -%}
Beore correction 368 lines after 97.
Now the HTML output is cleaner

I have not found any

kind regards
Seb


iBug:
If you're looking to make the output HTML more compact, either check out the [`compress` layout](https://mmistakes.github.io/minimal-mistakes/docs/layouts/#compress-layout) or use something like [jekyll-tidy](https://github.com/apsislabs/jekyll-tidy) like [I do](https://github.com/iBug/iBug-source/blob/a2efebc826994d3b6ef4a14b6d90a75ef0fe148c/Gemfile#L46).

This kind of blind search-and-replace is not going to be productive and will sooner or later introduce obscure bugs, especially when operating on non-HTML parts like JS and CSS.
So thank you for the idea, but I'm not going to accept this change.

PR #5233, opened 2025-05-17
feat: add configurability which share links to present per post #1 

cschwartz:

## Summary

This allows users to select (in _config.yml) via
```yaml
share_on:
  x: true
  facebook: true
  linkedin: true
  bluesky: true
```
which social share links are displayed per post.
Defaults are set to true, so that there should be no breakage for default installations.

Happy to work with your requirements and make changes as needed :)

## Context
No, I was not sure if there is an issue category for feature requests.
If this is not something you want in this project, we can still close the PR.

### DIFF
--- a/_includes/social-share.html
+++ b/_includes/social-share.html
@@ -1,19 +1,27 @@
 <section class="page__share">
   <h4 class="page__share-title">{{ site.data.ui-text[site.locale].share_on_label | default: "Share on" }}</h4>
 
+  {% if site.share_on.x | default: true %}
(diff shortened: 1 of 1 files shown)


iBug:
This might not work as intended. [Liquid documentation](https://shopify.github.io/liquid/filters/default/):

> default will show its value if the input is `nil`, `false`, or empty.

So an explicitly specified `site.share_on.x = false` *will* cause the button to show instead.

Test your code to show I'm wrong.

PR #5241, opened 2025-05-22
Preload Header Overlay Images

SocksTheWolf:

## Summary

Header images are the most important thing outside of content. While this could also be done in the general custom header file, it's probably best the theme is practicing best practices. Be responsive for all users out of the box.

Adds a liquid check to inject a preload `<link>` tag with the overlay_image if it's specified.

## Context

If the user has overlay images in their page header liquid, make it so we by default, preload those images immediately. This will make the page look responsive and clean.


iBug:
This PR is based on theoretical argument of presumed optimization, so I'm not considering this before concrete numbers prove non-trivial improvement.

PR #5368, opened 2025-11-16
Support resolving `ui-text` using `page.locale` when available

neveler:

## Summary

Support resolving `ui-text` using `page.locale` when available.

## Context

In multilingual setups, individual pages may specify their own `locale` value. This change enables the theme to use that value when selecting the appropriate `ui-text` entries, providing more precise control over per-page localization while remaining fully backward-compatible.


neveler:
> This looks good, why is it closed?

Thank you for reviewing! This PR was originally submitted to the project, but since I hadn't received any feedback for a long time, I decided to close it. However, if you think it can be merged, I’m happy to reopen it and discuss further. Please let me know if you need any modifications or additions from my side.

PR #5453, opened 2026-03-17
Respect custom parameters in yiq-contrast-color function

neveler:
This is a bug fix.

## Summary

The yiq-contrast-color function ignored its $dark and $light parameters and always returned global default variables instead.

This change ensures the function correctly uses the provided arguments, while still falling back to default values when not specified.


neveler:
https://github.com/timhettler/sass-yiq/blob/master/sass/module/_module.scss

PR #5477, opened 2026-04-27
Allow per-post or per-collection `date_format` values.

e28eta:
This is an enhancement or feature

## Summary

Update `_include/page__date.html` and `_include/page__meta.html` to read custom `date_format` values from the `page` / `document` object.

## Context

I wanted to add a `collection` to my site, and use a different date format for them (Month Year, omitting the day), and discovered it wasn't a built-in feature.

I think this is a reasonable approach, and seems in line with the recent addition of `locale`. My specific use case is using the `defaults` to avoid repeating it in every YAML front matter block:

```yaml
defaults:
  - scope:
      path: ""
      type: portfolio
    values:
      layout: single
      show_date: true
      date_format: "%B %Y"
```

### DIFF
--- a/_includes/page__date.html
+++ b/_includes/page__date.html
@@ -1,5 +1,5 @@
 {% assign locale = include.locale | default: site.locale %}
-{% assign date_format = site.date_format | default: "%B %-d, %Y" %}
+{% assign date_format = page.date_format | default: site.date_format | default: "%B %-d, %Y" %}
 {% if page.last_modified_at %}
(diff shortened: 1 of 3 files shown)


e28eta:
For what it's worth, I thought this was already a valid way to set the date format. Back in Dec 2024 I added `show_dates: true` and a custom `date_format` to the same block in config's defaults, and apparently never noticed that the date format wasn't what I'd asked for. I must have been skimming this section of the docs, and not realized it was supposed to be a top level key 


https://github.com/e28eta/e28eta.github.io/commit/6800dd007f06989fa973092e8afa9b3556cda26e

iBug (approved):

PR #5486, opened 2026-06-01
update `.full` style

neveler:

## Summary

Update `.full` style.

### DIFF
--- a/_sass/minimal-mistakes/_utilities.scss
+++ b/_sass/minimal-mistakes/_utilities.scss
@@ -174,7 +174,7 @@ $text-alignments: left, right, start, end, center, justify;
 
 .full {
   @include breakpoint($large) {
-    margin-inline-end: -1 * span(2.5 of 12) !important;
+    margin-inline-end: -1 * $right-sidebar-width-narrow !important;
   }
 }
 


iBug:
Can you elaborate on what this PR changes or fixes?

neveler:
> Can you elaborate on what this PR changes or fixes?

My understanding of .full is that it is intended to allow the element to overflow so that it can span both the content area and the right sidebar content. I’m wondering whether, since the variable right sidebar width narrow already exists, it would be better to use that directly. Additionally, using this variable would make the element’s fill area wider than the region covered by span.

iBug:
Can you provide screenshots for before/after the change? Right now it doesn't seem very obvious.

neveler:
> Can you provide screenshots for before/after the change? Right now it doesn't seem very obvious.

image

```html
<div class="full">
<img src="https://nginx.org/img/nginx_logo.svg" />
</div>

<div class="full-2">
<img src="https://nginx.org/img/nginx_logo.svg" />
</div>
```

```scss
.full-2 {
    @include breakpoint($large) {
        margin-inline-end: -1 * $right-sidebar-width-narrow !important;
    }
}

#main {
    outline: red solid 1px;
}

.full {
    outline: green solid 1px;
}

.full-2 {
    outline: orange solid 1px;
}
```

PR #5488, opened 2026-06-07
Use relative paths for README images

neveler:

## Summary

Move README to `docs` and update README image sources to use relative paths.


iBug:
There's a reason `README` is located at the root of the repository. Moving it breaks and hurts a lot of things.

neveler:
> There's a reason `README` is located at the root of the repository. Moving it breaks and hurts a lot of things.

It seems that the current build output contains static assets for the preview images of the newly added skins. We'd like to exclude these unnecessary assets from the final bundle.

iBug:
OK then. You may relocate the assets and update links, but please keep `README.md` at the repository root. The README file is the portfolio of the project and is not meant to be moved.

screenshot-layouts.png
iBug, line 1: Why are these two files deleted?

iBug (changes requested):

screenshot-layouts.png
iBug: Ah I got it. They're duplicated inside `docs/` directory.

PR #5494, opened 2026-06-19
Fix animations to pass Lighthouse

jnsiemer:
This is a bug fix.

## Summary

This PR sets all values that were `opacity: 0` before to `opacity: 0.01`.

## Context

I noticed that several MinimalMistakes pages have problems passing Google's Lighthouse tests, which is relevant for their Google search index. In my case, the test / simulation failed completely and returned an error code `NO_FCP`, which basically says that the page doesn't render anything while the page loads perfectly fine to the human eye.

Turns out that Lighthouse ignores elements that have `opacity: 0` at the start of their lifetime and therefore, doesn't recognise the "final" rendered page. Thus, I set all of the `opacity: 0` instances to `opacity: 0.01`, which should suffice to circumvent this issue.

### DIFF
--- a/_sass/minimal-mistakes/_animations.scss
+++ b/_sass/minimal-mistakes/_animations.scss
@@ -4,7 +4,7 @@
 
 @-webkit-keyframes intro {
   0% {
-    opacity: 0;
+    opacity: 0.01;
   }
   100% {
     opacity: 1;
@@ -13,7 +13,7 @@
 
 @keyframes intro {
   0% {
-    opacity: 0;
+    opacity: 0.01;
   }
   100% {
     opacity: 1;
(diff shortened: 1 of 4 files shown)


_sass/minimal-mistakes/_animations.scss
iBug, line 1: Blindly changing `opacity: 0` to `0.01` is not a good idea. Only this file can have their `opacity: 0` replaced and changes in all other files should be reverted.

iBug (changes requested):

_sass/minimal-mistakes/_animations.scss
iBug: > all other files should be reverted

It's not done yet.

iBug (changes requested):

iBug (approved):

